PR #24, opened 2021-11-25
feat: basic data migration from legacy RSD and setup of Postgres and …

ewan-escience:
…PostgREST in docker-compose

This is a basic setup that includes Postgres and PostgREST using docker-compose. It also includes a simple Java program that migrates a few fields from the legacy RSD software table. It can be used to further build on and to test e.g. Nuxt.js 3 with. 

Note that this is just an initial setup. We will later use better security practices related to the database and PostgREST.

### DIFF
--- a/data-migration/src/main/java/nl/research_software/migration/Main.java
+++ b/data-migration/src/main/java/nl/research_software/migration/Main.java
@@ -1,0 +1,69 @@
+package nl.research_software.migration;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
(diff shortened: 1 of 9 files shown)


data-migration/src/main/java/nl/research_software/migration/Main.java
```diff
@@ -0,0 +1,69 @@
+package nl.research_software.migration;
```
jmaassen, line 1: Can we use `nl.esciencecenter.rsd.migration` as a package here? 

The `nl.esciencecenter` part is standard for our java code.

jmaassen (changes requested):

PR #36, opened 2021-11-30
feat: upgrade to PostgREST 9, Postgres 14 and add timestamp fields an…

ewan-escience:
…d triggers to software table

### DIFF
--- a/database/002-create-software-table.sql
+++ b/database/002-create-software-table.sql
@@ -2,5 +2,32 @@ CREATE TABLE software (
 	id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
 	slug VARCHAR(100) UNIQUE NOT NULL,
 	brand_name VARCHAR(100) NOT NULL,
-	read_more VARCHAR
+	read_more VARCHAR,
+	created_at TIMESTAMP NOT NULL,
+	updated_at TIMESTAMP NOT NULL
 );
+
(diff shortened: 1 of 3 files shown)


dmijatovic:
@ewan-escience Looks good to me. It can be merged.

dmijatovic (approved):

database/002-create-software-table.sql
```diff
@@ -2,5 +2,32 @@ CREATE TABLE software (
...
-	read_more VARCHAR
+	read_more VARCHAR,
+	created_at TIMESTAMP NOT NULL,
+	updated_at TIMESTAMP NOT NULL
 );
+
+CREATE FUNCTION sanitise_insert() RETURNS TRIGGER LANGUAGE plpgsql as
+$$
+BEGIN
+	NEW.id = gen_random_uuid();
```
ctwhome, line 13: I like we are using the 'pgcrypto' extension, you could also use the "uuid-ossp" extension to generate UUIDs

ctwhome (commented):
Looks good.

ctwhome (approved):

ctwhome:
Please refer to the issue #31 to connect the PR

database/002-create-software-table.sql
ewan-escience: I don't think we use any extension, this function is built into Postgres since version 13: https://www.postgresql.org/docs/13/functions-uuid.html.

PR #45, opened 2021-12-06
Expand database tables

ewan-escience:
This will migrate most of the data from the legacy RSD to the SaaS version. Some final data still needs to be done, but this PR is already quite big.

Note that as more data is downloaded and processed, the script will take longer to run (on my computer up to 20 seconds).

Also PostgREST is exposed on port 3500 instead of 3000.


ctwhome (changes requested):
Can you add briefly in the README what are the parts connected to each other and the workflow? Also how to run the backend locally step by step?

dmijatovic (approved):

ctwhome (changes requested):
Two changes to make the development experience easier:

1 - Move the content of the README to the root README under the section `Running locally` and another different section with `Build`. They are two different concepts.
2 - First-time installation / DB reset use `docker-compose database init` (or similar), to serve the app without migration leave it with simply `docker compose up`

ctwhome (approved):
very clear explanation, thanks

jmaassen (approved):
Looks good! Go ahead and merge

PR #58, opened 2021-12-17
Migrate software index page

### LINKED ISSUE #55: Migrate software index page
# Software index page

Legacy software index page should be migrated to new RSD. There are some desired improvements 
- The filtering on the right side of legacy page should be improved. All filter values are shown by default. In case of lot of items the filtering items will expand on the right side of the page
- Filtering on the company name can be removed. In the new RSD the companies will have different role and it is expected that the filter will be less relevant

## Legacy software index page
image

## Concept updated page layout

Main suggestion is to move search, sort, filtering and navigation into the header of the page. This ensures that all navigation options are at the top of the page and accessible. The navigation section should be visible at any time (it should be fixed on the top of the page).

image

dmijatovic:
# Software index page (first itteraction)

Resolves #55 
Migrated most of the functionality of software page with desired improvements as described in #55.

How to test:

* remove old volumes: `docker-compose down --volumes`
* start new instance, migrate data and build: `docker compose up --build`
* open frontend at: http://localhost:3000
* navigate to Software page: http://localhost:3000/software
* type search term 'pharma': it should return 1 software
* try filtering on technologies using filter icon next to search bar
* when no match is found, basic `no content` message is shown (we plan to improve design later)

## Notes

* filtering on technologies is done by matching any of tags (logical OR). This is different from legacy version where software was require to have all selected values (logical AND). When selecting multiple technologies very few software meets the criteria.
* sorting using sort icon, next to filter icon, has ONLY one option enabled for now. Other options are not ready yet, as they require further DB migration of additional tables.
* `last updated message` in the card uses only updated_at information from the software table. See next point too
* `updated_at` date will be the date of data-migration. So, when testing it should be `1 hour ago` for all software (as you just created DB and migrated all data). We need to improve migration script to use original software values.
* current design might need futher improvements. Suggestions are welcome!

## Expected result 

image


ewan-escience (commented):
Consider filtering when typing in the search box after a small delay instead of having to press enter. Same when clicking on a tag for filtering.

dmijatovic:
@ewan-escience Thanks for the suggestions. I implemented these

PR #64, opened 2021-12-20
On error show no content message

### LINKED ISSUE #63: Software overview page doesn't work when the backend is not available

dmijatovic:
# On error show no content message

Fixes #63 

How to test:

* start only frontend without backend and visit the software page. You should see the website and simple no content message in the middle of the page.

image


ewan-escience (commented):
It works as described. However, the "No content" on the software page looks different from the one at the project page. Not a deal breaker though, as we didn't have project content yet. But this must eventually be consistent. 

Also please try to put a newline at the end of each file.

ctwhome:
we need to lint-fix the code before committing. @dmijatovic  did you find the way to format the code with eslint (+ prettier) for next?

README.md
```diff
@@ -12,6 +12,13 @@ The program can easily be built with `docker-compose`:
```
ctwhome, line 12: thanks for the docs

frontend/utils/getSoftware.ts
```diff
@@ -77,5 +76,6 @@ export async function getTagsWithCount(){
     }
   }catch(e:any){
     logger(`getTagsWithCount: ${e?.message}`,"error")
+    return []
   }
 }
```
ctwhome, line 81: I suggest to mode the external calls with fetching to a folder called API, utils can be confusing since is has a different meaning.

ctwhome (changes requested):

frontend/utils/getSoftware.ts
ctwhome: Grid and cards are pure CSS components that we can make with tailwindcss, I would prefer that

dmijatovic: > I suggest moving the external calls with fetching to a folder called 'api', utils can be confusing since is has a different meaning.

@ctwhome I think with Next calling the folder api might be confusing as it has api functionality. I would rather leave these function in utils, or maybe create folder data? Not sure.

dmijatovic:
> It works as described. However, the "No content" on the software page looks different from the one at the project page. Not a deal breaker though, as we didn't have project content yet. But this must eventually be consistent.
> 

Agree. The project "no-content" component will be the same as software "no-content" component.

> Also please try to put a newline at the end of each file.

Just runned default linter and it does not have rule for new line at the end of the file. I will look at the linter rules.

ctwhome:
We could make a nice illustration or similar, but for now it looks good to me, no errors!

frontend/utils/getSoftware.ts
dmijatovic: > Grid and cards are pure CSS components that we can make with tailwindcss, I would prefer that

Is replaced with tailwind classes

frontend/.eslintrc.json
```diff
@@ -1,6 +1,12 @@
...
   "rules":{
-    "no-console": "warn"
+    "no-console": "warn",
+    "eol-last": "warn",
+    "quotes": ["warn", "double"],
+    "semi":["warn", "never"],
+    "object-curly-spacing": ["warn", "never"],
+    "array-bracket-spacing": ["warn", "never"],
+    "no-trailing-spaces": "warn"
   }
```
ewan-escience, line 11: Please add the rule `"indent": ["error", "tab"]`

ewan-escience (changes requested):
See the comment

frontend/.eslintrc.json
dmijatovic: > Please add the rule `"indent": ["error", "tab"]`

Can we investigate this first and do it in separate PR? The default for frontend development is to use spaces. I have tried this couple of years ago and eventually reverted back to spaces because all defaults are set to spaces. In my experience it will require continuous linting and corrections.

ctwhome (approved):

ewan-escience (approved):

PR #65, opened 2021-12-21
Admin role

ewan-escience:
# Admin role

* Anonymous users can only see published entries
* Admins can see and edit everything

How to test: 

* Copy the extra lines from `frontend/.env.local.example` to `frontend/.env.production.local` and `frontend/.env.local`
* Delete everything: `docker-compose down --volumes`
* Build: `docker-compose build database data-migration`
* Start: `docker-compose up database backend data-migration`
* The data-migration script should after a while finish successfully
* After the data migration is done, the following page should give an empty array: http://localhost:3500/software?select=slug&is_published=eq.false
* Create a jwt at with payload claim `"role": "rsd_admin"` signed with the secret in `frontend/.env.production.local` using HS256
* Make a request to the same address as above but with a bearer authorization header with the jwt you created. You should see a bunch of results.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/data-migration/src/main/java/nl/esciencecenter/rsd/migration/Main.java
+++ b/data-migration/src/main/java/nl/esciencecenter/rsd/migration/Main.java
@@ -1,22 +1,31 @@
 package nl.esciencecenter.rsd.migration;
 
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.algorithms.Algorithm;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 
+import java.io.FileReader;
 import java.io.IOException;
 import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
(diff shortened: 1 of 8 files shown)


dmijatovic (approved):
Works as intended, including expiration date/time. Very nice!

data-migration/src/main/java/nl/esciencecenter/rsd/migration/Main.java
```diff
@@ -1,31 +1,48 @@
...
 import java.util.Arrays;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Properties;
 
 public class Main {
 
+	static final long TEN_MINUTES_IN_MILLISECONDS = 600_000L; // 10 * 60 * 1000
```
ctwhome, line 26: I kind of miss a bit of documentation per function of methods.

ewan-escience: I'm not sure what you mean. Each method has a descriptive name and at places where I do something that's not obvious I added some comments to explain.

PR #67, opened 2022-01-04
Auth fix account creation

ewan-escience:
# Auth fix account creation

* The redirect url for SURFconext is now read from the env files instead of being hardcoded
* Fix accounts not being created because no rsd_admin token was sent to PostgREST
* Tokens are now sent as cookies
* Add endpoint to refresh tokens

How to test:

* Build the authentication service. 
* Start the database, backend and authentication services. 
* Go to http://localhost:7000/login/surfconext and follow the link. After signing in, you should see a JWT on the screen. The same JWT should be in a cookie with name "rsd_token". This cookie should have the Secure and HttpOnly values to true and the SameSite value to Lax. 
* Then go to http://localhost:7000/refresh. Your cookie should be replaced with a new one. The token should have the same content, except that is it valid again for one more hour (and the signature was updated)

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


dmijatovic (approved):
It works as described

PR #68, opened 2022-01-04
60 software item page - first part

dmijatovic:
# Software item page: part 1

This is first part of the software item page design. There is already lot of code created and it seem to be good moment to make PR and have these sections integrated.

Sections covered in this PR and the remarks:
- Introduction: design and back-end integration completed.
- Get started: the chart with commits and other info about commits is not implemented yet. It is still unclear if we can scrape this data 'on-demand' from Github or to migrate commit data from legacy.
- Cite this software: design and backend integration is completed. There are some deviations from legacy design: 
   - The notification 'popup' (bottom-left) is used when copying DOI instead of change of button label
   - The style of the buttons is slightly different
- About section: the design is completely implemented with a slight deviation from the legacy page (some icons are different and the accordion is used instead of the button for read more section). The read more section has now an maximum height of 30 rem. If the content is larger the scrollbar is applied to this section. The programming languages are not implemented. We consider scraping this information 'real-time' from Github. This is not implemented yet.

How to test:
- remove old database: `docker-compose down --volumes`
- build application: `docker-compose build`
- start application: `docker-compose up` 
- navigate to one of the software item pages, for example http://localhost:3000/software/kernel-tuner
- mentions and contributor counts should be identical to ones in the legacy RSD
- get started link should take you to a same location as the same button in the legacy RSD
- DOI's, when changing the version, should be identical to ones displayed in the legacy RSD. You can copy DOI to the clipboard and notification will appear in the left bottom corner
- you can download the citation file, after selecting desired format. The content of the citation file should be identical to one from the legacy RSD.
- in the about section, the title and the bullets should have identical content as in legacy RSD. Read more section should have identical content as in the legacy RSD under read more button.
- the tags section should have same entries as in the legacy RSD. Note! in the new version the tags are alphabetically ordered. If software does not have tags, the message `No tags available` is shown
- programming languages section has dummy values Test 1, Test 2, Test 3 (info is not migrated yet and we consider different approach)
- license section should have identical values as in legacy RSD
- source code section should have identical value(s) as in legacy RSD

You should see a page similar to image below:
image


dmijatovic:
> On the layout: the page uses the full width of my screen, which doesn't look good in my opinion. The legacy RSD seems to use a max width, consider using that,

Thanks for the remark, we will optimize container size for different screens later. I will merge it as it is for now,

frontend/pages/software/[slug]/index.tsx
```diff
@@ -16,17 +17,27 @@ import {
...
+  {
+    slug: string, software: SoftwareItem, citationInfo: SoftwareCitationInfo, tagsInfo: Tag[], licenseInfo: License[],
+    softwareIntroCounts: ContributorMentionCount
+  }) {
   const [options, setSnackbar] = useState(snackbarDefaults)
 
-  console.log('licenseInfo...', licenseInfo)
+  if (!software?.brand_name){
+    return (
+      <ContentInTheMiddle>
```
Comment (ctwhome) [inline frontend/pages/software/[slug]/index.tsx:36]: please remove <ContentInTheMiddle> component and replace with CSS styling.

ctwhome (commented):
Styled components should rely on css classes

frontend/pages/software/[slug]/index.tsx
Comment (dmijatovic) [reply frontend/pages/software/[slug]/index.tsx:36]: Thanks. I'll have look at it at the next PR on the software page.

PR #138, opened 2022-02-16
133 ux new software page

### LINKED ISSUE #124: Add instant feedback on duplicated slug when creating a new software page
The form needs to give instant feedback when typing like it does when writing the short description. 

*   If the number of requests is a concern, we can "debounce" the calls to the API. we can use lodash.debounce or an [alternative](https://dmitripavlutin.com/react-throttle-debounce/)
*   Let's also edit the slug if the editor wants
*   Remove the error alert, and shot it in its context, underneath the slug input as the error message

### LINKED ISSUE #126: New software modal opens a new route
When clicking on a new software page, the application redirects to a new route that opens a modal form. 

The logic behind a modal is to keep the context-awareness of any page, letting know the user hasn't gone anywhere else and his/her chances are not lost. When moving to a new page we need to make sure the user can save any modifications made.

To not change the functionality of the router, I would remove the modal and show the information on a page

### LINKED ISSUE #133: Improve UX on creating new page

ctwhome:
# Improve UX on creating new page

Closes #133
Closes #124 
Closes #126  

*   [x] Move modal to page
*   [x] Apply changes on inputs
*   [x] Make slug editable by user

image

*   [ ] Update documentation
*   [x] Tests


dmijatovic:
@ewan-escience and @jmaassen The page is updated. Can you review new version please?

jmaassen (approved):
Looks good.

Maybe a small (possibly future) UI improvement could be the way in which the URL is shown. Now it shows: 

https://localhost/software/   slug 

Where only the "slug" part can be changed and the spaces in front don't count. This last bit is a bit weird? Could we remove the spaces and see if that also works?

PR #246, opened 2022-05-02
234 login options

dmijatovic:
# Flexible login options

Closes #234 

* RSD instance can be defined through env. variable RSD_AUTH_PROVIDERS to offer the login with SURFconext, Helmholtz AAI or both
* When one provider is defined the Sign in button redirects directly (as in the past)
* When more than one provider is defined, the modal with the list of the providers is shown. The order of providers in the list is identical to the order of values in the env variable. The values of the providers should be semicolon (;) separated.
* When RSD_AUTH_PROVIDERS is missing, the default provider is SURFconext
* The default values are specified in the api functions for each provider in case that env variable is missing (please check if defined defaults are correct). The name of the provider, used in the button, is also specified there.

How to test:
* Extend you .env file with RSD_AUTH_PROVIDERS, see env.example file
* Specify one or both providers (not case sensitive, semicolon separated)
   *  RSD_AUTH_PROVIDERS=surfconext;helmholtzaai
   *  RSD_AUTH_PROVIDERS=surfconext
   *  RSD_AUTH_PROVIDERS=helmholtzaai
* `docker-compose build`: to build front-end code
* `docker-compose up`: to start
* Try without, with one and both providers in the env file
* Testing login with both providers. I was able to login with SURFconext but not with Helmholtz AAI (using ORCID) on localhost. The Java code logs error about the organisation

## Suggestion for further improvements

- Use well-know endpoint url to obtain the token url instead of AUTH_HELMHOLTZAAI_TOKEN_URL, AUTH_SURFCONEXT_TOKEN_URL. I would advice this improvement in the auth module (Java)
- ~~Investigate why Helmholtz AAI fails with ORCID. I was able to login on remote version with ORCID, however the localhost does not work~~  **Resolved by using flag HELMHOLTZAAI_ALLOW_EXTERNAL_USERS**

## Example multiple providers

image

PR Checklist:

*   [ ] Update documentation
*   [x] Tests


dmijatovic:
> @dmijatovic You may test logging in (for Helmholtz) with your GitHub account, it should work now. It worked for me on localhost. But for me also ORCID worked (after registration process and mail verification).

@ruester Did the login works for you in this branch/PR? I am able to login with ORCID on dev version (old codebase) but in this PR it does not work for me. My first assumption is that it could be something with the localhost return path. Otherwise I might have not implemented everything correctly in this PR.

ewan-escience:
I cannot login through GitHub both on localhost and on the VM, in both cases because no code is transmitted to the backend.

ruester:
I've tested it on this branch an both GitHub and ORCID worked. It may be important that you use a private browsing tab or remove all cookie data. It sometimes did not work for me (as I was developing the integration of the Helmholtz AAI) if I logged in with GitHub, logout and then login again with ORCID. I guessed it has something to do with cookies or so as a private browsing tab always worked then...

ruester (approved):
Thanks, works like expected :smiley: :+1:

ewan-escience:
I tried it in a private window, I get the following query parameters back: `{error=[server_error], error_description=[Unexpected server error]}`

ruester:
> I tried it in a private window, I get the following query parameters back: `{error=[server_error], error_description=[Unexpected server error]}`

Tested with GitHub? Could you also test ORCID or Google? Those should definitely work.. Also don't forget to set `HELMHOLTZAAI_ALLOW_EXTERNAL_USERS=true` as described in the `.env.example`

ewan-escience:
Yes, tested with GitHub. I don't have Google or ORCID. :) Thanks for reminding me, but `HELMHOLTZAAI_ALLOW_EXTERNAL_USERS` is not relevant here, as the error occurs before this is used. I'll wait until you hear about the possible bug affecting my account. :)

dmijatovic:
> HELMHOLTZAAI_ALLOW_EXTERNAL_USERS

@ruester Thanks for the tip!!! I forgot to set external users flag ;-). I am able to login with ORCID on localhost, same as on dev. I think this works as intended.

Do you want to update auth code to use well-known endpoint instead of env variable in this PR or prefer a separate issue?
With the new setup one might consider renaming env variables. The NEXT_PUBLIC part of the name is not required (it was at the beginning). The naming is completely free, if you update the provider part of the api code.

ruester:
So now some environment variables are not needed anymore? Which one? Or is it just a renaming thing because those variables are not used in the frontend anymore, but still needed in the backend?
Is the SURFconext class from the auth module also affected? If so we may open a new PR for this so that @ewan-escience and I make the changes for both auth classes.
As I saw no changes in the SURFconext env variables I was not aware that we may change something now in the backend...

dmijatovic:
@ruester, @ewan-escience  Maybe is better to create new issue and discuss the changes? This PR works in the current state properly. 

I saw possible improvement in using well-know endpoint instead of AUTH_HELMHOLTZAAI_TOKEN_URL and AUTH_SURFCONEXT_TOKEN_URL in the auth module. Then these env. variables could be removed. In addition, some other renaming could be done if desired but it is not necessary. All suggestions I made are optional.

ewan-escience:
Let's do that in a separate branch.

ctwhome (approved):

cmeessen (approved):
Looking good! Login via Helmholtz->GitHub doesn't work for me either.

I have one minor suggestion: I think it would make sense to add a comment about the default sign in option if `RSD_AUTH_PROVIDERS` in `.env.example`.

cmeessen:
I need to correct myself: GitHub login **is working**. I did not notice that I received a confirmation mail that required some action by clicking on a confirmation link. After that, it worked.

PR #253, opened 2022-05-08
252 edit project team

### LINKED ISSUE #164: Improve ORCID affiliations info
When searching for contributors using ORCID api, the affiliations returned are array of all person affiliations, most likely in alphabetical order. It would be more useful to have this list in the historical order and select current/last affiliation as value for the RSD affiliation.

image

### LINKED ISSUE #252: Edit project - Team
This implementation should be similar to Contributors section of software page

dmijatovic:
# Edit project team

Closes #252
Closes #164

* Project maintainer add/edit/delete team members using find/add options
* We search in RSD and ORCID database
* If person extracted from ORCID has more than one affiliation we show them in drop-down. User still can type in completely different affiliation. Same solution is implemented the edit contributors modal.

How to test:
* `docker-compose down --volumes`: remove old
* `docker-compose build`: to build solution
* `docker-compose up`: to start
* data-migration is optional
* login using any of test users, eg. professor1
* create new project
* add team members using search on the right side of the screen, the section should look like in the example below
* when using information from ORCID, which might provide multiple affiliation values, first affiliation from the array is used and the other values can be selected from dropdown. To make clear that multiple options are present, the dropdown will open on focus. The help text is updated to state "Select or type in...". Beside selecting the dropdown options the user can type in any text value.

image

## Affiliation from ORCID example

image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -36,6 +36,7 @@ RSD_AUTH_URL=http://auth:7000
 # consumed by services: frontend (api/fe)
 # provide a list of supported OpenID auth providers
 # the values should be separated by semicolon (;)
+# if env value is not provided default provider is set to be SURFCONEXT
 RSD_AUTH_PROVIDERS=SURFCONEXT;HELMHOLTZAAI
 
 # SURFCONEXT - TEST ENVIRONMENT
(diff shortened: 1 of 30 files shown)


ruester (commented):
Works like expected. One improvement could be that the text field ("Find or add team member") is cleared after the member was added, since the user probably not add one person twice.

ruester (approved):

PR #256, opened 2022-05-09
[Feat] embed-pages inside iframes

ctwhome:
# Ability to embed RSD pages within an iFrame

Fixes #250

*   [x] Be able to detect the URL param `?embed` to display or not the footer and header sections within the iFrame
*   [x] Display back button when embed option
*   [x] Works for Software pages and Project Pages (for now)

Note: this is a working example of functionality that works with any type of HTML

### Use example:

```html
<h2>Check our Software - Demo Embed pages</h2>
<iframe src="http://localhost/software?embed" title="RSD Software" width="100%" height="500" frameBorder="0"></iframe>

<h2>Check our Projects - Demo Embed pages</h2>
<iframe src="http://localhost/projects?embed" title="RSD Software" width="100%" height="500" frameBorder="0"></iframe>
```

How to test:

*   Run `make start`
*   open: http://localhost/embed\_example.html
*   you should be able to pick any software, search and go back to the main grid view after visiting a software of project

PR Checklist:

*   [x] Update documentation
*   [ ] Tests


ctwhome:
NextJS dynamic links with LINK override the parameter when changing between pages within the iframe. We need to find a way to keep the embed property alive. Probably with React context. 

We also need to take a detailed look at how navigation works between pages when navigating through the iframe. Menu breadcrumbs and back buttons work very well and they are mostly implemented in the application already.

ctwhome:
I am planning to work on the documentation for this feature soon and we are good to test and merge

ewan-escience (approved):
Works for me. If testing without running the frontend separately, don't forget to remove the `:3000` parts from the iframes.

ctwhome:
@dmijatovic yes I would do the same. So this was a POC to see the embedding part working, which it does and that's really good. But a second step, I would add a new parameter in the route to only show software and projects from a specific organization. Then we could actually embed it in our corporate website directly.

Issue #201, opened 2022-04-06
Pull authors from zenodo


ctwhome:
Message from Peter 

Hi, I was adding my software to the RSD, but wondered whether filling some of the information could not have been done automatically. After having it filled in for the citation.cff and on Zenodo, I needed to fill in things like the contributors again in the RSD. Would be nice if this could be somehow synchronized maybe?

cmeessen:
We have this one planned in research-software-directory/Helmholtz#9

PR #262, opened 2022-05-11
Auto fill in keywords

ruester:
# Auto fill keywords

Now same as the contributors we can get the keywords from the DOI.
See #201 and https://github.com/research-software-directory/Helmholtz/issues/9

@dmijatovic you may have to improve the UI again (you can add the commits in this PR if you want :wink:), for example the page scrolls to top after keywords were loaded which is not ideal, but I don't know how to prevent this nicely since we would need to add anchors to the headings (?)

I want also to do the same for the license, since we can get it from DataCite too, but this will be another PR. But those three things (contributors, keywords, license) are all we can get from a DOI automatically for now, I think...

* import the keywords from the DataCite API if a DOI was given

How to test:

* create a new software entry and enter a DOI which has keywords in the data (for example: 10.6084/m9.figshare.90828 or 10.5880/riesgos.2021.003)
* click on "Import Keywords", keywords should be added, but not saved yet, no duplicate keywords are added
* a DOI without keyword information does not add any keywords (example: 10.5281/zenodo.2585848)

PR Checklist:

*   [ ] Update documentation
*   [X] Tests


cmeessen:
Thank you, it works really well. I noticed it states that concept DOIs are used, but also non-concept DOIs are working (e.g. 10.5880/GFZ.6.1.2017.001).

Also, it seems the DataCite subject keywords contain HTML, e.g.

```
https://api.datacite.org/dois/10.5880/riesgos.2021.003

..
{
"subject": "EARTH SCIENCE SERVICES &gt; DATA ANALYSIS AND VISUALIZATION",
"subjectScheme": "NASA/GCMD Earth Science Keywords"
},
...
```

It should be show as *EARTH SCIENCE SERVICES > DATA MANAGEMENT/DATA HANDLING*. I think trusting the response and parsing everything is too risky, but we may replace `&gt;` with `>`?

Another question is how to handle nested keywords? Does it make sense to split them or use only those with the most levels? From the same response:

```
{
"subject": "EARTH SCIENCE SERVICES &gt; WEB SERVICES",
"subjectScheme": "NASA/GCMD Earth Science Keywords"
},
{
"subject": "EARTH SCIENCE SERVICES &gt; WEB SERVICES &gt; DATA PROCESSING SERVICES",
"subjectScheme": "NASA/GCMD Earth Science Keywords"
}
```

Adding both keywords seems redundant. In that case adding the second one would make most sense from my point of view.

cmeessen (changes requested):
See comment above :)

ruester:
> Thank you, it works really well. I noticed it states that concept DOIs are used, but also non-concept DOIs are working (e.g. 10.5880/GFZ.6.1.2017.001).

As long as the scrapers still need a concept DOI I did not want to change it for now.

> Also, it seems the DataCite subject keywords contain HTML, e.g.

Yes it seems like the DataCite API just converts the XML to JSON but does not decode the content. 

> It should be show as _EARTH SCIENCE SERVICES > DATA MANAGEMENT/DATA HANDLING_. I think trusting the response and parsing everything is too risky, but we may replace `&gt;` with `>`?

We should not replace content returned from the API, the API should instead fix the values :wink: 

> Another question is how to handle nested keywords? Does it make sense to split them or use only those with the most levels? 

It is still possible to remove unwanted keywords before saving the software entry. As those keywords come from GFZ we should speak with the Data Services there.. They are using GCMD Science Keywords which are structured this way, I guess

ruester:
>     * multilevel keywords: we do not support multilevel keywords. The question is how easy is to extract single (relevant) level from this. I saw that these keywords do have specific type. I did tested some of our software and we seem to have "flat/simple" keywords. I would suggest that we might ignore these "custom" types of keywords (I see that in api these are called subject and have in addition subjectScheme property).

exactly.. the keywords are just "free text" and mainly useful for filtering or search engines...

Thanks for your commits, lets continue at #267 than :+1: 
(Next time I will push directly to this repo for a better PR workflow :wink: )

PR #272, opened 2022-05-16
270 correct platform type migration

### LINKED ISSUE #270: Set the correct repo url platform type when migrating
When migrating from the legacy RSD, we should detect which repository urls are from GitHub and which are from other platforms. Currently, only one entry has a non-GitHub repo url, namely the entry with slug `magnumpi` (it is not published so you cannot see it).

ewan-escience:
# Detect code platform when migrating

* The data migration script does a few simple checks to see if the code platform of a repo url is from GitHub, GitLab, Bitbucket or something else

How to test:

* Start with an empty database:
	* `docker-compose down --volumes && docker-compose up --scale scrapers=0`
* Build and run data migration:
	* `cd data-migration && docker-compose build && docker-compose up`
* Run the following query and check that the results are correct (there is only one row to check at the time of writing, which is a GitLab repo):
	`SELECT * FROM repository_url WHERE url NOT ILIKE '%github%';`
* Also check that the rest is indeed GitHub:
	`SELECT * FROM repository_url WHERE url ILIKE '%github%';`

Closes #270

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/data-migration/docker-compose.yml
+++ b/data-migration/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     build:
       context: ../
       dockerfile: ./data-migration/Dockerfile
-    image: rsd/data-migration:0.0.12
+    image: rsd/data-migration:0.0.13
     env_file:
       # use main env file
       - ../.env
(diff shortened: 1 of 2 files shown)


dmijatovic (approved):
Works as expected!

PR #275, opened 2022-05-18
Draft: Auto fill licenses from DOI

ruester:
# Auto fill licenses from DOI

Fixes https://github.com/research-software-directory/Helmholtz/issues/9

* fetch licenses from DOI information (using DataCite API)

How to test:

* create a software entry and fill in a DOI containing license information, click on "Import licenses", the licenses from the DOI should then be added
* using a DOI without license information should not add licenses and show a message instead

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


dmijatovic:
@ruester Nice work! The select component for licenses is not designed to be driven in the same way as keywords component. Initially we planed licenses to be "closed list" from which user can select an option. From the feedback we receive from users the closed list of spdx licenses might not be sufficient. The license component might need some refactoring anyway, as we want to "drive" values by code (as import function). I need to put bit more thinking into this and talk to others about exact approach we would like to use in this case (we also have scrapers scraping the license info from the repository).

For your easier understanding of the code, important portion of the code that drives the form values is using [react-hook-form](https://react-hook-form.com/api). Reading through the api could help you better grasp the code. 

To sum up, this PR requires some more discussion and thinking before we can implement it. I expect to work on it next week.

PR #278, opened 2022-05-18
264 simple test login

### LINKED ISSUE #264: Implement simple test login
We want to test live SURFconext login for eScience Center employees.

Because we have other clients using dev to test features and need some type of test login we cannot replace current SURFconext test idp login with live one, instead we need to have additional login options in dev environment:

@ewan-escience proposed to create simple test provider that will accept any use name and provide a valid token in order to test all feature of "normal" rsd user.

# Next steps

- create additional test auth provider in authentication module
- define it in env variables
- adapt frontend part to support test provider

ewan-escience:
# Local login

* Allow for local account creation when the value `LOCAL` is part of env variable `RSD_AUTH_PROVIDERS`
* These account are *not secure* and should be used for testing purposes only
* On the auth module, the endpoints are now disabled if the corresponding values are not a part of `RSD_AUTH_PROVIDERS`; in the case of a missing `RSD_AUTH_PROVIDERS`, SURFconext is enabled

How to test:
* Add `LOCAL` to the env variable `RSD_AUTH_PROVIDERS` (see also the new `.env.example`)
* Build and run
* Click on Sign in
* Click on Local account
* Fill in a username of your choice
* After some redirects, you should now be logged in

Closes #264

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,10 +33,11 @@ POSTGREST_URL=http://backend:3500
 # .env.local: http://localhost/auth, .env: http://auth:7000
 RSD_AUTH_URL=http://auth:7000
 
-# consumed by services: frontend (api/fe)
+# consumed by services: authentication, frontend (api/fe)
 # provide a list of supported OpenID auth providers
 # the values should be separated by semicolon (;)
 # if env value is not provided default provider is set to be SURFCONEXT
+# if you add the value "LOCAL", then local accounts are enables, USE THIS FOR TESTING PURPOSES ONLY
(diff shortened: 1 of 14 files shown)


dmijatovic (approved):
Nice work! You are natural FE dev talent :1st_place_medal:

ruester (approved):
Works as expected

cmeessen (approved):
:+1:

PR #291, opened 2022-05-25
271 fix scraping releases

### LINKED ISSUE #271: Scraping releases sets the updated_at for the software entry
In the `software` table, there is the column `releases_scraped_at`. This field is used to prioritise which software should have its releases scraped next. However, when any change is made to the table, the field `updated_at` is set to the current time. This field is used to show on the website when an entry was last updated. 

We should prevent the scraping of releases to influence this `updated_at` field. A solution could be to place it in a different table. Maybe the `repository_url` could be used, even though releases have nothing to do with this table. Another solution could be to have a separate table just for this purpose.

ewan-escience:
# Fix scraping releases updates the timestamp when the software page was updated

* Move the column `releases_scraped_at` from `software` to `release`
* Create a function in the database that allows (again) for sorting on `releases_scraped_at`
* The `releases_scraped_at` is updated first, so that even if an error occurs later, the entry is pushed back to the end of the scraping order
* A fix that sets `is_citable` and `latest_schema_dot_org` for the corresponding entry only instead of all entries
* Set `NULL` instead of the empty string for `latest_schema_dot_org` when it is missing

How to test:
* The database structure has changed, clear and build:
	* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up`
* Run data migration
* Scraping releases should work, you can speed it up by running `docker-compose exec scrapers python3 releases.py` a few times
* Check (`SELECT * FROM "release";`) that different releases have different values for `latest_schema_dot_org`

Closes #271

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/002-create-software-table.sql
+++ b/database/002-create-software-table.sql
@@ -15,7 +15,6 @@ CREATE TABLE software (
 	is_featured BOOLEAN DEFAULT FALSE NOT NULL,
 	is_published BOOLEAN DEFAULT FALSE NOT NULL,
 	short_statement VARCHAR(300),
-	releases_scraped_at TIMESTAMP,
 	created_at TIMESTAMP NOT NULL,
 	updated_at TIMESTAMP NOT NULL
 );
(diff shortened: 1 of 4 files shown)


dmijatovic (approved):
Works as expected.

PR #294, opened 2022-05-30
Add license and copyright statements

### LINKED ISSUE #293: Adding license information to all files
Following my question from the last weekly, I would like to open this issue to discuss how license information could be enhanced in the repository.

I think that a very nice approach is used by the [REUSE Specification](https://reuse.software/). They provide a tool that can automatically add license and copyright information. It can be implemented via [pre-commit hooks](https://reuse.software/dev/#pre-commit-hook) to make it easy for the developers. Also, they have a linter that can be integrated in a [pipeline](https://reuse.software/dev/#ci) to check conformity.

This is an example for a license header from their page:

```
# SPDX-FileCopyrightText: 2016, 2018-2019 Jane Doe <jane@example.com>
# SPDX-FileCopyrightText: 2019 Example Company
#
# SPDX-License-Identifier: GPL-3.0-or-later
```

There is one drawback as far as I see: REUSE requires that licenses are placed in a `LICENSES` folder in the root of the directory. It makes sense in a way that here you can put multiple license texts, e.g. if texts or images have a `CC-BY` license. But in doing that, GitHub does not automatically detect the license anymore. But this may not be the case in the future anymore, since there are some [requests](https://github.com/licensee/licensee/issues/490) in the licensee repository that ask for REUSE support and the maintainer is open for that.

REUSE supports a machine readable [debian copyright](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0) file under `/.reuse/dep5`. That makes it easy to initially apply a license and copyright to all the files. However, as far as I could see, this cannot be automated, and it becomes tedious to take care about multiple copyright holders for individual files.

Finally, REUSE is also supported by the best practices badge mentioned in #153, and adding license/copyright info on a per file level is part of the "Gold" level.

/edit: my suggestion is to use a script to add the initial copyright statements. This would include:

* Copyright to all files for the NLeSC
* Add copyright of all authors that edited the specific files
* Add license statement
* Add a section to CONTRIBUTING and the documentation about how to configure your git to automatically add this info

@jmaassen @dmijatovic @ewan-escience @ctwhome would you be happy with this solution? I'd be happy to do the implementation for this.

cmeessen:
# Add license and copyright statements

Fixes #293 

This is the working branch for adding license and copyright information to all files.

Proposed changes and additions:

* adds a `.mailmap` to unify the used commit authorships (there are some commits with different names for the same contributors) @dmijatovic @ewan-escience @jmaassen @ctwhome @Normo please check whether the replacement I chose are the right ones. The format is `Unified name <unified@mail> Actual commit name <actualcommitmail>`.
* adds a new CI job that checks the repository for REUSE compliance upon merge requests
* adds instructions on how to (optionally) use git pre-commit hooks to automatically add headers on every commit

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen:
I'll pick this up and apply it to the whole repo.

cmeessen:
I just added instructions on how to use the pre-commit hooks for git. Right now, the script just lives in `CONTRIBUTING.md`. We could also create a file so one would only have to copy-paste the file into `.git/hooks`, but I didn't know where to put it.

What would still have to be changed in the script is which license will be assigned to which file type.

jmaassen:
> I just added instructions on how to use the pre-commit hooks for git. Right now, the script just lives in `CONTRIBUTING.md`. We could also create a file so one would only have to copy-paste the file into `.git/hooks`, but I didn't know where to put it.
> 
> What would still have to be changed in the script is which license will be assigned to which file type.

I'm still a bit hesitant to ask everyone to install a pre-commit hook, since it may fail (and thus fail the commit) if it encounters a situation it cannot handle... I think I would be in favor of a setup which only complains if the license information is missing somewhere, plus a (manually triggered) script which helps you add it where needed. That way, one can intervene if the script fails, or when a new combination of file type / license is encountered. 

Would that also be an option?

cmeessen:
> I'm still a bit hesitant to ask everyone to install a pre-commit hook, since it may fail (and thus fail the commit) if it encounters a situation it cannot handle

That's a fair point. How about we soften the description in `CONTRIBUTING.md`, making it clear that the script is optional and not mandatory. In case the script fails, I would suggest to add another terminal output mentioning that to commit anyway one should use `git commit -n`.

A check for existing license information is performed anyways upon every pull request, so we would detect missing headers the latest at this point.

cmeessen:
@jmaassen I updated `CONTRIBUTING.md` and the script to address your points. There is some more detailed information about how to add the headers, how to check them manually and automatically, and I emphasised that the pre-commit hook is optional. Also, the pre-commit hook now continues committing if an error occurs.

dmijatovic:
@cmeessen Yes, I read the same in the docs, there is no way to exclude files from checks in the action. I would like to propose to run this action only on PR not on push to main.

cmeessen:
Another idea would be to use a custom pipeline that only checks the files that will be committed. Thinking about it, this would be my preferred solution.

dmijatovic (approved):
Let's give it a try and see how it will work in practice!

cmeessen:
> Let's give it a try and see how it will work in practice!

Fine by me. I do not have anything to push to this branch anymore. If you are ready, you can merge.

PR #296, opened 2022-05-31
Add list of projects for the user

### LINKED ISSUE #283: User profile - Projects
Show the list of the projects the user is maintainer of. The design and the approach should be similar to organisation/projects section. 
The basic scaffold is done. Use the rpc created for this view (projects_by_maintainer) and the components used on organisation pages.

## Starting point
image

## Desired implementation based on organisations pages

image

ruester:
# Add list of projects for the user

Fixes #283

* add an overview of all projects to the user page for which the user is the maintainer of
* fix some minor logging messages
* remove an unused file (accidentally added some time ago?)

How to test:

* add some projects with a logged in user and you should see them in the projects list on the user page
* adding projects with another user should only display the projects for which the user is maintainer of

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


frontend/components/user/project/index.tsx
```diff
@@ -1,26 +1,73 @@
...
-    setTimeout(() => {
-      setLoading(false)
-    },1000)
-  },[])
+    if (count && loading === false) {
+      setCount(count)
+    }
+  }, [count, loading, setCount])
+
+  // do not use loader for now
```
ruester, line 30: I removed the loading like it is done in `~/components/user/software/index.tsx` too

frontend/components/user/project/useUserProjects.tsx
```diff
@@ -0,0 +1,121 @@
...
+export async function getProjectsForMaintainer(
+  {searchFor, page, rows, session}: UserProjectsProp
+) {
+  try {
+    // baseUrl
+    let url = `/api/v1/rpc/projects_by_maintainer?maintainer=eq.${session?.user?.account}&order=is_published.desc,title`
+
+    // search
+    if (searchFor) {
+      url += `&or=(title.ilike.*${searchFor}*, subtitle.ilike.*${searchFor}*))`
```
ruester, line 30: this has potential for not working with special characters as it is not URL encoded, but it was also done in `getSoftwareForMaintainer` this way... also I don't know how much of a security issue it can be, as I could search for something like `)&or=(title.eq.project)#` and change the overall query a bit...
I do not know so much about PostgREST so I cannot decide on this with absolute certainty :thinking:

cmeessen (approved):
Very nice, thank you. Works as described.

frontend/components/user/project/useUserProjects.tsx
dmijatovic: @ruester good point! We can encode URL requests. I did that in some requests [(/utils/postgrestUrl.ts)](https://github.com/research-software-directory/RSD-as-a-service/blob/main/frontend/utils/postgrestUrl.ts) but clearly not in all. So we can definitely improve on that. I am not worried about query manipulation, we have LRS in Postgress and the entries you are not allowed to see will not be shown no matter the query spec. I'll have more in detail look at your PR this afternoon.

dmijatovic (approved):
Fantastic work!

frontend/components/user/project/index.tsx
dmijatovic: Yes, sure

frontend/components/user/project/useUserProjects.tsx
dmijatovic: You can make this improvement in this PR or later. I am ok with both

PR #338, opened 2022-06-15
Add global navigation to app header

ctwhome:
Fixes #

* Adds a global search to navigate between software, projects and organizations      

How to test:

*

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ctwhome:
It need the connection to the API to fetch the real results

ewan-escience:
There is a visual bug when using the search box on the edit pages only, see the screenshot:

image

database/100-create-api-views.sql
```diff
@@ -977,3 +977,22 @@ BEGIN
...
 $$;
+
+
+--     Global search
+--     Create
+CREATE FUNCTION global_search() RETURNS TABLE(
+    name VARCHAR,
+    slug VARCHAR,
+    source TEXT
+    ) LANGUAGE plpgsql STABLE SECURITY DEFINER AS
```
ewan-escience, line 988: Please remove `SECURITY DEFINER`. This will prevent leaking data that should not be shown. You can remove the `WHERE is_published=TRUE` clauses then. A nice consequence is that you can also search your own unpublished software. And RSD admins can search all software, even unpublished entries.

And please use tabs instead of spaces for indentation. :)

ewan-escience (changes requested):
Except for the bug I posted earlier and the comment in this review, it works very nicely!

ctwhome:
I see that material ui is having a bigger z-index defined for the sticky
header.

I like the idea of see also unpublished stuff if you are the maintainer.
Probably i should add as well an indication of "unpublished" in the search,
and the redirect to the edit mode.

Tabs vs spaces, aren't we using by default spaces in the Frontend? At least
i think we have it like that set on eslint, but i need to double check

dmijatovic:
> I see that material ui is having a bigger z-index defined for the sticky header.

Yes, maybe be can set sticky header to z-9 instead of 10?

I have suggestion about the position. I am not in favor of menu moving on focus, maybe we can place the searchbar on the other side of menu items or on the right side as bellow?

image

frontend/components/GlobalSearchAutocomplete/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+        />
+
+        {isOpen &&
+          <div
+            className="shadow-xl absolute top-[50px] w-full left-0  bg-white text-black py-2 rounded">
+            {!hasResults && <div className="px-4 py-3 font-normal bg-gray-200 mb-2 "><span
+              className="animate-pulse">No results...</span></div>}
+            {searchResults.map((item, index) =>
+              <div key={index}
+                   className={`${selected === index && 'bg-[#09A1E3] text-white'} flex gap-2 p-2 cursor-pointer transition justify-between items-center`}
```
dmijatovic, line 135: This approach generates "false" classes, maybe just use basic ternary `selected === index ? 'bg-[#09A1E3] text-white':''`

ewan-escience:
I only meant to use tabs in the SQL file. :)

ctwhome:
Thanks for the suggestions!

The position and animation of the searchbar is intentional, so the
searchbar comes inmediatelly after the logo giving more priority to the
search than to the menus.  See github.com 's approach which has the same
use case.

I think the search inour at the end of the navigation bar looks nice but it
falls into a second category in terms of what the users sees first.

The animation should give a feeling of dinamism and fluently. It shouldn't
be a problem because the eye fixation remains on the search when activate
Ly clicking there and not in the menu items

frontend/components/GlobalSearchAutocomplete/index.tsx
ctwhome: improved, thanks!

ewan-escience (approved):
Nice work!

PR #340, opened 2022-06-16
Add possibility to define a whitelist for user logins

ruester:
# User Whitelist

See https://github.com/research-software-directory/Helmholtz/issues/4

This helps us for a first productive deployment of the RSD in order to evaluate, present, test and get further feedback.
So that only some defined test users can login to the RSD and create entries and we not ending up having many "test 123" entries in the list :wink: 

* add configuration variable to define user mail addresses which are allowed to access the RSD
* this should only be considered as a temporary implementation as long as we have no role management
* whitelist is **not** activated for the local login

How to test:

* set the env variable `RSD_AUTH_USER_MAIL_WHITELIST` to the mail address of your login account
* start the containers, you should be able to login only with this account
* unset the env variable again and every user should login again
* you can also define multiple mail addresses with a comma-separated list in the env value

PR Checklist:

*   [X] Update documentation
*   [X] Tests


.env.example
```diff
@@ -46,6 +46,11 @@ RSD_AUTH_URL=http://auth:7000
 # if you add the value "LOCAL", then local accounts are enables, USE THIS FOR TESTING PURPOSES ONLY
 RSD_AUTH_PROVIDERS=SURFCONEXT;HELMHOLTZAAI
 
+# Defina a comma-separated list of user email adresses which are allowed to
+# access the RSD.
```
cmeessen, line 50:
```suggestion
# login to the RSD. If the variable is left empty, or is not defined, all users
# will be allowed to login.
```

Here is one suggestion for a little more explanation.

cmeessen (approved):
Thank you :+1: works like a charm. Let's see what the others are thinking about this.

.env.example
```diff
@@ -46,6 +46,11 @@ RSD_AUTH_URL=http://auth:7000
 # if you add the value "LOCAL", then local accounts are enables, USE THIS FOR TESTING PURPOSES ONLY
 RSD_AUTH_PROVIDERS=SURFCONEXT;HELMHOLTZAAI
 
+# Defina a comma-separated list of user email adresses which are allowed to
```
dmijatovic, line 49: The RSD_AUTH_PROVIDERS env variable uses semicolon (;) as delimiter. I would prefer to use same delimiter in both cases (actually in all places if possible).

dmijatovic (approved):
It works as expected. If possible, use same delimiter.

.env.example
ewan-escience: I agree, let's keep it consistent. :)

ruester: now implemented with semicolon :+1:

ruester:
I've added your suggestions and rebased to main again, should be ready to merge now :wink:

PR #341, opened 2022-06-17
minor: auto-create .env.local

cmeessen:
# Create .env.local with Makefile

Here is a minor suggestion to improve the `Makefile` by @ctwhome:

* Use the `Makefile` to auto-generate the file `frontend/.env.local` from `.env`
* Uses `sed` to replace `POSTGREST_URL`

How to test:

* delete (or make a backup) of your `frontend/.env.local`
* `make dev-frontend`
* Should now tell you that it creates `frontend/.env.local` and then starts `yarn dev`
* shutdown the dev server
* `make dev-frontend`
* starts the server without recreating `.env.local`
* shutdown the dev server again
* edit the `.env`, e.g. `echo "# just a comment" >> .env`
* `make dev-frontend` should detect the change and regenerate `.env.local`
* shutdown the dev server one more time
* edit the `.env.local`, e.g. `echo "# another comment" >> frontend/.env.local`
* `make dev-frontend` should start the dev server without changing `frontend/.env.local`

When starting from scratch `make dev-frontend` will fail and tell that it requires `.env`

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+# SPDX-FileCopyrightText: 2022 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+# SPDX-FileCopyrightText: 2022 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 # SPDX-FileCopyrightText: 2022 Jesús García Gonzalez (Netherlands eScience Center) <j.g.gonzalez@esciencecenter.nl>
 # SPDX-FileCopyrightText: 2022 Netherlands eScience Center
 #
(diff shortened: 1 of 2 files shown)


ctwhome (approved):
It generates perfectly fine the .env.local. One more step automated! many thanks!

Makefile
```diff
@@ -33,5 +35,10 @@ down:
 dev-docs:
 	cd documentation && yarn dev
 
-dev-frontend:
+frontend/.env.local: .env
+	@echo "Creating frontend/.env.local"
+	cp .env frontend/.env.local
+	sed -i 's/POSTGREST_URL=http:\/\/backend:3500/POSTGREST_URL=http:\/\/localhost\/api\/v1/g' frontend/.env.local
```
ruester, line 41: this is not enough I guess, since RSD_AUTH_URL is also different for the frontend

cmeessen: True, it is stated in the `.env.example`, but not in the `frontend/README.md` to change it. I did not change it and do not have issues running the frontend in dev mode. @dmijatovic is this variable required to be changed?

cmeessen: I just rebased to the current master, and force pushed the changes to this branch. Also, the `Makefile` now searches and replaces `RSD_AUTH_URL`.

**However**: I tested it again, and there are no issues running the frontend in dev mode if `RSD_AUTH_URL=http://auth:7000`. I could login via all three providers.

PR #430, opened 2022-07-22
feat: enable docker-compose project names

### LINKED ISSUE #429: Support docker-compose project names with Makefile
When handling this repo, and the HIFIS RSD repo locally, both repositories always use the same containers. This sometimes leads to incompatibilities, because the content of one RSD may not be compatible to use in the other RSD.

I suggest we add a new optional environment variable to `.env`. If this one is set, we can use it in the makefile to define the docker-compose project name with `docker-compose --project-name STRING`

cmeessen:
# Enable docker-compose project names

Fixes #429 

Allow using project names for docker-compose, so the same containers and volumes will not be used for different versions of the RSD. Example: you have the origin repo in `~/git/RSD-as-a-service`, and a fork in `~/git/fork-project/RSD-as-a-service/`, they will share the same volumes, networks and containers (because `container_name` is defined). In case there are differences in the DB structure or elsewhere, this might cause some troubles.

This approach makes use of the `COMPOSE_PROJECT_NAME` environment variable (see [docs](https://docs.docker.com/compose/reference/envvars/#compose_project_name=)). Since the property `container_name` is specified for each container in the `docker-compose.yml`, this property would override `COMPOSE_PROJECT_NAME`. This is why I removed the `container_name`. The relations between the containers are not affected, their names refer to the id's in the yml.

* Adds a new environment variable `COMPOSE_PROJECT_NAME="rsd"` to `.env.example`
* removes `container_name` properties in `docker-compose.yml`

How to test:

* `docker-compose --down` to clear networks and volumes (optional)
* leave `COMPOSE_PROJECT_NAME` undefined or `COMPOSE_PROJECT_NAME=""`
  * `docker-compose up`
  * will create containers and networks with default names, in this case with prefix `rsd-as-a-service` (=directory name)
* set `COMPOSE_PROJECT_NAME="rsd"`
  * `docker-compose up`
  * will create containers named `rsd-<container>-<#>`, and volumes/networks with prefix `rsd_<..>`

PR Checklist:

*   [x] Update documentation (`.env.example`)
*   [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,14 @@
 
 # You also need to obtain/generate missing secrets
 
+# ---- DOCKER PROJECT SETTINGS ----------
+
+# Define this variable, if you are running different versions of the RSD, in
+# order to define the docker project name. If you leave this empty, docker will
+# automatically name the containers.
+COMPOSE_PROJECT_NAME="rsd"
+
+
 # ---- PUBLIC ENV VARIABLES -------------
 
 # postgresql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,6 @@ version: "3.0"
 
 services:
   database:
-    container_name: database
     build: ./database
     image: rsd/database:1.0.1
     ports:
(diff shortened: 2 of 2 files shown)


dmijatovic (approved):
Nice work!

PR #439, opened 2022-07-25
Add support for website tracking via Matomo

### LINKED ISSUE #435: Add support for Matomo
For webstatistics, we are using a self hosted instance of [matomo](https://matomo.org/). We suggest to implement a feature equivalent to the hotjar integration. If a matomo ID is given in the `.env`, the matomo integration should be activated.

ruester:
# Add support for website tracking via Matomo

Fixes #435

* similar to Hotjar: now Matomo is also supported for website tracking

How to test:

* set the env variables MATOMO_URL and MATOMO_ID
* you should then see a request from the frontend to `<YOUR_MATOMO_URL>/piwik.php?...&idsite=<YOUR_ID>`
* if one (or both) of the variables is not set no request to matomo is send

PR Checklist:

*   [X] Update documentation
*   [ ] Tests


cmeessen (approved):
I can confirm that the matomo code appears in the frontend, however I cannot verify that it actually connects.

cmeessen:
@dmijatovic would you like to review too or can we merge this?

dmijatovic (approved):
It looks good to me on the first sight. I assume you tested it ;-)

ruester:
> however I cannot verify that it actually connects.

you can only see the request in the developer console for example... but matomo itself does not track the visits if the request are not sent from the registered domain, so don't expect page visits to appear in matomo for now.. we have to deploy it to production in order for it to work properly

cmeessen:
@ruester thanks, we will have to wait until production is live anyways, because Matomo does not track GFZ IP addresses (but those are the only ones allowed to access yet).

PR #472, opened 2022-08-03
464 case sensitive doi

### LINKED ISSUE #460: Makefile/Readme - differing/non existing make actions
In the Makefile: `make start | install | dev | down`
In the Readme: `make up | install | dev | down | local`

`local` needs to be defined. I guess `start` refers to `up`.

### LINKED ISSUE #464: project pages: case sensitivity of entered DOIs
if I enter output with DOI in capitalized letters (e.g., 10.4230/DagMan.6.1.1), RSD does not see the existing entry and tries to add it from another source (e.g., DataCite). This leads to an error as shown in the screenshot
Screenshot from 2022-07-29 13-54-48

If I enter a title or its DOI in lowercase letters (in the same example, 10.4230/dagman.6.1.1), the existing RSD entry is found.

This could potentially be problematic to users, because i just copy pasted DOI the way it was written in the article.

ewan-escience:
# Make DOI's case insensitive

* Columns in the database that store a DOI now use the case insensitive variable type `CITEXT` (see https://www.postgresql.org/docs/14/citext.html)
* Fix the mentions scraper to be case insensitive with respect to DOI's
* This update breaks the data-migration script, so we deprecate it and update the documentation

How to test:
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up`
* Add a software page, add a mention with doi `10.4230/DagMan.6.1.1`. The mention should be found in Datacite.
* Add another software page, add a mention with doi `10.4230/dagman.6.1.1` (note that it's the same DOI but with different casing). It should find this mention in the RSD instead of Datacite. You should be able to add this mention without an error.
* Use DBeaver or something comparable to change the DOI of mention in the database to `10.4230/DagMan.6.1.1`. Run the scraper (`docker-compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.doi.MainMentions`). It should finish without an error, and the DOI of the mention will be changed back to small letters (`10.4230/dagman.6.1.1`)

Closes #464
Closes #460

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 # SPDX-FileCopyrightText: 2022 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+# SPDX-FileCopyrightText: 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 # SPDX-FileCopyrightText: 2022 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 # SPDX-FileCopyrightText: 2022 Jesús García Gonzalez (Netherlands eScience Center) <j.g.gonzalez@esciencecenter.nl>
 # SPDX-FileCopyrightText: 2022 Netherlands eScience Center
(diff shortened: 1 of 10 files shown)


ewan-escience:
In the production database, there are duplicate DOI's for mentions, i.e. DOI's that are the same but have different casing. I wrote a script that can merge them, which can be found [here](https://github.com/research-software-directory/RSD-production/blob/main/database-migration/duplicate-dois/merge-duplicate-dois.sql). This script was tested on the dev vm.

cmeessen (approved):
I can confirm, everything works as descirbed :+1:

jmaassen (approved):
Works like a charm!

Issue #392, opened 2022-07-11
Implement cookie notification


jiskattema:
A quick look shows that the website places 2 cookies; however, there is no cookie notification or banner.
Given that the cookies are purely functional, permission to store is probably not necessary. However, showing a message that we use functional cookies only, and nothing else (ie. google analytics..?) could be useful.

jmaassen:
Good suggestion, we'll look into this

jmaassen:
We may have to check with legal what the rules are on this?

cmeessen:
Since with #435 we integrated support for Matomo, we would also need a cookie notification banner where the user has the choice to accept or decline tracking. Ideally, based on the environemnt variables, the banner would automatically detect what to show.

In case of Matomo, or other tracking tools, there also needs to be the option to withdraw the consent elswhere. This could either be done on a dedicated privacy policy page (with more custom text), or on a dedicated "Cookie settings" page that might be referenced in the footer. In my opinion, the cookie settings page would be more transparent and more easily accessible than the solution in the privacy policy page. On the other hand, this may potentially clutter the footer. Do you have an opinion here?

@ruester is currently implementing the consent banner (ready to be merged here soon), and we are now thinking about where to place the withdrawal-option.

jmaassen:
Part of PR #488

PR #410, opened 2022-07-14
Support for custom themes via volume mounts and settings.json

dmijatovic:
# Add support for loading different themes at runtime

**This is WIP, created to explore the possibility of applying Helmholtz styles and typography**. 

Next steps are:
- Helmholtz to examine the approach and try to apply all their styles using the code in this PR
- RSD team to provide support and explain how the solution works

## Changes proposed in this pull request
* The theme approach is adapted to load different, predefined, themes at the runtime using volume mounts of frontend image.
* The homepage is loaded based on the host value provided in the mounted settings.json file. If host is not provided in the settings.json the default rsd homepage is loaded.
* If `settings.json` is not provided via volume mount the default settings file is loaded (public/data/settings.json)
* if `index.css` file with font imports is not provided via volume mount the default index.css file is used (public/styles/index.css)
* When using default RSD settings one does not need to provide volume mounts in the frontend service

## Open issues
* `Helmholtz footer`: different footer with additional image, different contact email and custom links to be added. My preference is to extend settings.json with these additional properties

```json
{
  ...other props
   "footer":{
      "links":[{"label":"Link label", "url":"https://some-url.com"}],
      "logo_svg":"<svg></svg>",
      "contact": "some.email@contact.com",
   }
}
```

* `Dark/light theme switch`: the basic infrastructure is created but the complete implementation involves additional steps
   * integrating theme switch in tailwind portion
   * integrating theme switch in tailwind markdown plugin. It uses it's own css variables
   * integrating theme switch in material-ui themes. Investigate use of css variables

How to test:
* `docker-compose build` to build the solution
* create `hmz` folder in the deployment directory. Note I have added hmz folder in the .gitignore file
* Copy all folder and files from the rsd folder into hmz folder. Provide your values in the index.css, settings.json files and upload your custom images used on your homepage. * 
* change values in docker-compose frontend service to:

```yaml
  # frontend definitions example to mount custom theme
  # add volumes as in this example to load index.css and settings.json
  frontend:
    ...other definitions
    volumes:
      # mount local styles folder into styles folder of frontend app (index.css file is the access point)
      - ./deployment/hmz/styles:/app/public/styles
      # mount custom settings (theme colors and typography definitions) into data folder of frontend app
      - ./deployment/hmz/data:/app/public/data
      # mount custom images used on custom home page (for helmholtz only)
      - ./deployment/hmz/images:/app/public/images
    ...other definitions
```
[...]

PR #474, opened 2022-08-03
Add Matomo cookie banner

ruester:
# Add cookie banner for Matomo tracking

Fixes https://github.com/hifis-net/RSD-as-a-service/issues/37
See also #392

See [Matomo tracking-consent documentation](https://developer.matomo.org/guides/tracking-consent) for technical details.

* add a banner on all pages to ask the user for consent on being tracked (+cookies) from Matomo
* for this we use the library [react-cookie-consent](https://www.npmjs.com/package/react-cookie-consent) which offers a solution for a customizable cookie banner
* on a separate page the user is able to revoke or accept the tracking consent again

How to test:

* use a private browser tab or remove all your cookies for localhost
* opening any page of the RSD will show an overlay asking the user to consent/decline to the tracking+cookies of Matomo
* no requests on first page load are send out to Matomo
* accepting the tracking lead to requests being send out to the Matomo server
* by clicking on "decline" no requests are send to the Matomo ever and until this cookie is valid
* at the page `/cookies` the user gets some information about cookies in general and is able to revoke the matomo tracking consent again, each of them having the effect of stopping or starting the tracking of Matomo

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -3,72 +3,37 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# Based on Next official example for minimal image size (standalone approach)
-# https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile
-# it requires output: 'standalone' in next.config.js
-
-# ----------------------------------------
-# 1. Install dependencies only when needed
-# ----------------------------------------
-FROM node:18.5-buster-slim AS deps
-
-WORKDIR /app
-
-# copy
-COPY package.json yarn.lock ./
-# install
-RUN yarn install --frozen-lockfile --silent
-
-# ----------------------------------------
-# 2. Build image
-# ----------------------------------------
-FROM node:18.5-buster-slim AS builder
-
-WORKDIR /app
-
-# copy node_module from deps image
-COPY --from=deps /app/node_modules ./node_modules
-# copy all files from repo (ones not defined in .dockerignore)
-COPY . .
+FROM node:18.5-buster-slim
 
 # Next.js collects completely anonymous telemetry data about general usage.
 # Learn more here: https://nextjs.org/telemetry
 # Uncomment the following line in case you want to disable telemetry.
 ENV NEXT_TELEMETRY_DISABLED 1
 
+WORKDIR /app
+
+# copy
+COPY package.json yarn.lock ./
+
+# install
+RUN yarn install --frozen-lockfile --silent
+
+# copy all files from repo (ones not defined in .dockerignore)
+COPY . .
+
 # test solution
 RUN yarn test
+
+ENV NODE_ENV production
 
 # build the solution
 RUN yarn build
 
-# ----------------------------------------
-# 3. Production image (standalone mode)
-# ----------------------------------------
-FROM node:18.5-buster-slim AS runner
-
-# optional install updates
-# RUN apt-get upgrade -y
-
-WORKDIR /app
-
-ENV NODE_ENV production
-
 RUN addgroup --system --gid 1001 nodejs
 RUN adduser --system --uid 1001 nextjs
-
-# You only need to copy next.config.js if you are NOT using the default configuration
-# COPY --from=builder /app/next.config.js ./
-COPY --from=builder /app/public ./public
-COPY --from=builder /app/package.json ./package.json
-
-# Using standalone output to reduce image size
-# https://nextjs.org/docs/advanced-features/output-file-tracing
-COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
-COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
 
 USER nextjs
 
 EXPOSE 3000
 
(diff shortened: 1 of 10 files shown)


ruester:
@dmijatovic This is what I have so far. Its quite a mess.. mostly due to the lack of my react/nextjs skills :sweat_smile: 

I also wanted to add a link in the footer to the cookie information site, but I will wait for a decision on how to integrate the "revoke feature" first (https://github.com/research-software-directory/RSD-as-a-service/issues/392#issuecomment-1203808123)

cmeessen (commented):
Looks like it works, thank you. I get some warnings in the console:

* on the frontpage, no cookies exist yet
* Click "Reject"
* `There was an error setting cookie `mtm_consent_removed`. Please check domain and path.`
* Is this because of the mismatch between `localhost` and `MATOMO_URL`?
* now I open `localhost/cookies` and click on accept, which prints the following warning:
`
Cookie “cookie_consent” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`

ruester:
There is currently also a problem for environment variables in the production build. I cannot get it to work so that the function components have the value for `MATOMO_URL` and `MATOMO_ID`... Its strange that it works for `_document.tsx` but not for `_app.tsx` or `cookies.tsx`.
It seems like those variables get deleted with some step, here I recorded a strange behaviour when reloading the page a few times:

https://user-images.githubusercontent.com/1287885/182875888-1130814a-06aa-4ba6-9ee1-5f8637a49750.mov

All the documentations I could find state that it should work. I also tried `publicRuntimeConfig` and `NEXT_PUBLIC_...`

@dmijatovic Could you help me getting the environment variables `MATOMO_*` to work also in production build?

ruester:
It seems to be a problem with the standalone mode for the production build: https://github.com/vercel/next.js/issues/38119
But if I use the latest version of next (12.2.4) then I run into problems with jest: https://github.com/aelbore/esbuild-jest/issues/26

I've now found a solution with using `publicRuntimeConfig` and without the standalone mode, but this should only be considered a temporary workaround and to prove that something is not working as expected (at least according to all the documentation about environment variables in nextjs/react I've read :wink:)

dmijatovic:
@ruester I started looking at this issue. Is there any specific reason for adding 2 additional dependencies: react-cookie-consent and react-use-cookie? Maybe we could solve it with dependencies we already have material ui, cookie.js lib and custom js?

Concerning the env variables. Accessing env variables from functional components (frontend only) is limited for security reasons. NEXT_PUBLIC...variables are available at build time only which is not sufficient for this case. AFAIK we can access env variables in the following ways:
- using getServerSideProps function defined at the page level, which unfortunately is not supported in _app.tsx. 
- by making a request to node api as done with auth env variables
- by loading mounted settings.json (also an fetch request) as I did it in [themes PR](https://github.com/research-software-directory/RSD-as-a-service/pull/410).

I have no experience with the approach you tried, [publicRuntimeConfig](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) but the documentation also points out that this approach only works with the server side pages (when page is using getServerSideProps) as far as I understand it.

To sum up, I stared looking at this issue and will come back with suggested approach after I make some tests. I am not familiar with Matomo so I will need to look at the documentation. Maybe you can get me at speed? How Matomo knows that user has given the consent for monitoring? Does cookie need to have specific name/values in it?

ruester:
> To sum up, I stared looking at this issue and will come back with suggested approach after I make some tests. I am not familiar with Matomo so I will need to look at the documentation. Maybe you can get me at speed? How Matomo knows that user has given the consent for monitoring? Does cookie need to have specific name/values in it?

Many thanks for having a look at this. I thinks the approach with the request to the node API is the way to go here, I did not thought of that :see_no_evil: 
I can go ahead, if you want, and change my current pull request to use the API to get the environment variable, but this may also slow down the first page load quite a bit?

Matomo sets specific cookies values so it knows if the user has given consent, see here: https://developer.matomo.org/guides/tracking-consent

You can test if Matomo is activated (or not) by reloading the page and then look at the developer console for a request to the given Matomo URL (or by looking at the cookie values in the developer console).

ruester:
> @ruester I started looking at this issue. Is there any specific reason for adding 2 additional dependencies: react-cookie-consent and react-use-cookie? Maybe we could solve it with dependencies we already have material ui, cookie.js lib and custom js?

Those may be left overs from my testing.. I think we do not need the react-use-cookie library for now :+1:
But the react-cookie-consent library was quite helpful since the cookie banner was easy to set up with this...

ruester:
Ah sorry.. just saw the new PR now (https://github.com/research-software-directory/RSD-as-a-service/pull/488)

PR #479, opened 2022-08-16
Improve visibility of pagination for grids

### LINKED ISSUE #446: Users tend to think there is only one page in the software grid
We have had multiple independent responses from testers that they initially thought, there is only one page in the software grid, because there are no arrows at the bottom of the page. They did not see the arrow in the upper right corner at first sight.

We have two suggestions to solve this:

* lazy load more software when reaching the end of the page
* add a "next page" button at the end of the page

/cc @dmijatovic

ruester:
# Add separate pagination component at bottom of grids

Fixes #446

We may change the look of the pagination or we may also hide it completely it if there are only 12 or less entries to show (so there is no next page)...

* add an additional pagination component at the bottom of each grid view (software, projects, organisations)

image

How to test:

* build the feature branch and add some entries (software, projects, organisations) to the DB
* the grids should have a pagination component at the bottom allowing the user to select a different page

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/pages/organisations/index.tsx
+++ b/frontend/pages/organisations/index.tsx
@@ -8,6 +8,7 @@ import Head from 'next/head'
 import {GetServerSidePropsContext} from 'next/types'
 import {useRouter} from 'next/router'
 import TablePagination from '@mui/material/TablePagination'
+import Pagination from '@mui/material/Pagination'
 
 import {app} from '../../config/app'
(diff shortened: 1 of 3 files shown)


dmijatovic (approved):
Nice work! Maybe we could hide the pagination (at the bottom) when there is only one page, but that can be done later (if there is a need for that functionality).

PR #486, opened 2022-08-25
433 team members

### LINKED ISSUE #433: Edit project - team members (feedback Rena)
# Adding team members

1. It does not offer to add new person when there is a match in ORCID but it is not same person
2. It does not support search by ORCID. Sometimes search returns lot of persons with the same name. It is easier to provide ORCID instead of name.

## Example Jan Jansen
image

## Note
The similar feature should be provided in contributors section of edit software page.

ewan-escience:
# Improve adding contributors and team members

* When entering a (syntactically correct) ORCID in the search field for contributors or team members, both the RSD and ORCID databases will be searched for on the ORCID field only
* It is now always possible to add a contributor or team member, even when results with a matching name have been found

How to test:
* `docker-compose build frontend && docker-compose up`
* Login, add a software page
* Search for an orcid, e.g. `0000-0003-4803-7411`. Only one result from the ORCID database should show up. Add this person as contributor. Search again with this ORCID, two results should show up (one from ORCID, one from the RSD)
* Search for a full name with results, e.g. `Jan Jansen`. Even though matches are found, it is still possible to manually add a contributor with this name
* Repeat this for team members for projects

Closes #433

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,7 +85,7 @@ services:
       # dockerfile to use for build
       dockerfile: Dockerfile
     # update version number to corespond to frontend/package.json
-    image: rsd/frontend:1.2.2
+    image: rsd/frontend:1.2.3
     environment:
       # it uses values from .env file
       - POSTGREST_URL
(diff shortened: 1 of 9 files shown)


dmijatovic (approved):
Excellent work! I have removed one legacy bug that became visible after implementation of this new feature.

PR #488, opened 2022-08-26
392 cookie consent

### LINKED ISSUE #392: Implement cookie notification

dmijatovic:
# Matomo cookie consent

Closes #392
Closes #474 

* Add support for cookie consent when MATOMO_ID provided as env variable
* Enable user to accept or decline anonymous data collection
* Enable user to change the value on cookies page
* App footer layout: 
   * contact email moved to left side
   * eScience logo is removed from the footer (it is modified and not identical with official logo)
* Unused invitations component: left/right margin removed

How to test:
* `docker-compose build frontend` to build frontend
* provide MATOMO_ID and MATOMO_URL values in .env file, see .env.example
* `docker-comopse up` to start
* on any page except /cookies you will see popup at the bottom
image
* If you, somehow manage to navigate to cookies page before using accept/decliine buttons, the page will open with off value by default. 
* The cookies page looks like this. You can accept/decline anonymous data collection.
image
* you can remove cookies and reload the page and the cookie consent dialog should appear again.

## Note
Beside testing the functioning of cookie consent components, please review used texts in the consent message and on the cookies page. It would be great if we can further improve the text used.

PR Checklist:

*   [ ] Update documentation
*   [x] Tests


ruester:
This looks great, thanks for all your work!

I have one issue: if I click on the cookie site link in the banner the switch for the consent is set to on (like a default value). So I have given consent just by entering the webpage which is not as expected. Could we change it to a default value for the switch to "off"?

ewan-escience:
It would be nice of this PR would also: 

1. add the Matomo env variables to `deployment/docker-compose.yml` (they are already in the top level `docker-compose.yml`)
2. add the [Matomo SPA code snippets](https://developer.matomo.org/guides/spa-tracking)

dmijatovic:
@ruester I changed the default value to off. In addition I removed the link to cookies page in the consent dialog to avoid setting any default value. However if the user somehow reaches the cookies page before giving consent the default value will be set to off (false).

ruester (approved):

jmaassen (approved):
Looks good. I've sent you a new text for the cookie page.

dmijatovic:
@ewan-escience, @ruester , @jmaassen I have implemented customUrl changes to be pushed to matomo analytics when appropriate. 

Few things I discovered:
- When you have **do not track** settings enabled in your browser the stats will not be passed, even if you have given consent. I think we should double check and confirm this.
- After disabling do not track in my browser (Chrome) I was able to push stats to Matomo instance from the localhost too. Maybe we can block this for the production site and create additional site for development where this is allowed?
- We do need to manually push route changes performed by next router, which is similar to SPA approach in documentation
- Search terms from software, project and organisation page are reported now. This is due to route change and matomo been able to detect search param in the url. One observation: if one search term is shown multiple times in subsequent order in the reporting it means that the user is paginating through the results.

ewan-escience (approved):

PR #506, opened 2022-09-05
359 Custom theme, footer logo, contact email and links

### LINKED ISSUE #359: Theme provider
Ability to support different themes, for eScienceCenter but also the rest of the organizations (Helmholtz).

Tasks: 
- [x] Unify theming between Tailwindcss and MUI
- [ ] Be able to switch dynamically in runtime between dark and light theme
- [ ] Document how to build themes
- [ ] Remove doubled themes from MUI

### LINKED ISSUE #374: Replace hard colors from the application  with the new theme variables
Across the application are written colors. 
Replace classes like `text-black` with `text-base-100` 
allowing to have different themes in the application.

### LINKED ISSUE #438: Pages administration: allow links to external pages
For convenience, we will not host the data protection declaration by ourselves, but will instead link to the declaration of the hosting institution (i.e. GFZ). It would be great if, instead of a complete markdown page, we could use the new administration frontend to add such links.

dmijatovic:
# Custom home page for Helmholtz and theme provider using settings.json.

Custom settings can be uploaded using volume mounts, see README in deployment folder.

Closes #359 
Closes #374 
Closes #410 
Closes #438

## Changes proposed in this pull request
* The theme approach is adapted to load different, predefined, themes at the runtime using volume mounts of frontend image.
* The homepage is loaded based on the host value provided in the mounted settings.json file. If host is not provided in the settings.json the default rsd homepage is loaded.
* If `settings.json` is not provided via volume mount the default settings file is loaded (public/data/settings.json)
* if `index.css` file with font imports is not provided via volume mount the default index.css file is used (public/styles/index.css)
* When using default RSD settings one does not need to provide volume mounts in the frontend service
* Footer links can be added using settings.json
* Footer logo and contact email are defined in settings.json

How to test:
* delete the directories `frontend/{node_modules,.next}` if they exist because they will be recreated by the frontend docker development container (and with the ownership of the docker system user, which most likely will be `root`, so changing the directory ownership with `chown` should also be possible instead of deleting them)
* `docker-compose build` to build the solution
* create `hmz` folder in the deployment directory. Note I have added hmz folder in the .gitignore file
* Copy all folder and files from the rsd folder into hmz folder. Provide your values in the index.css, settings.json files and upload your custom images used on your homepage. * 
* change values in docker-compose frontend service to:

```yaml
  # frontend definitions example to mount custom theme
  # add volumes as in this example to load index.css and settings.json
  frontend:
    ...other definitions
    volumes:
      # mount local styles folder into styles folder of frontend app (index.css file is the access point)
      - ./deployment/hmz/styles:/app/public/styles
      # mount custom settings (theme colors and typography definitions) into data folder of frontend app
      - ./deployment/hmz/data:/app/public/data
      # mount custom images used on custom home page (for helmholtz only)
      - ./deployment/hmz/images:/app/public/images
    ...other definitions
```
PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


dmijatovic:
@ruester and @ctwhome I created new PR which is continuation of #410. Custom theme implementation. I will close #410 in favor of this new PR

docker-compose.yml
```diff
@@ -85,7 +85,7 @@ services:
       # dockerfile to use for build
       dockerfile: Dockerfile
     # update version number to corespond to frontend/package.json
-    image: rsd/frontend:1.3.2
+    image: rsd/frontend:1.3.0
```
ruester, line 88:
```suggestion
    image: rsd/frontend:1.4.0
```

ruester:
> create hmz folder in the deployment directory. Note I have added hmz folder in the .gitignore file

Is it planned that we add our custom theme to your repository (research-software-directory)? Or should we keep our customization in our hifis-net repository? If so the `.gitignore` in `/deployment` should be changed, so that it does not ignore the `hmz` directory. It makes it difficult for developers to commit their custom contents to the deployment directory (need to use `--force` when adding a file)...

frontend/pages/index.tsx
```diff
@@ -5,271 +5,35 @@
...
-        <div className="mt-20">
-          <AppFooter/>
-        </div>
-      </div>
-    </div>
-  )
+export default function Home({counts}: HomeProps) {
+  const {host} = useRsdSettings()
+  // console.log('host...', host)
+  if (host && host.name.toLowerCase() === 'helmholtz') {
```
ruester, line 20: maybe this could be improved if you can set the component name dynamically like:

```react
if (host) {
  const compName = capitalize(host) + 'Home'
  <Component name="{compName}" />
}
```
pseudo-code due to missing react knowledge :sweat_smile: 
such a solution is possible in Vue (https://forum.vuejs.org/t/render-components-with-dynamic-names/14153/4), so I would assume this is also possible in React :wink: 

this way you would never have to touch this part of the index-component for other themes added in the future...

dmijatovic:
> Is it planned that we add our custom theme to your repository (research-software-directory)? Or should we keep our customization in our hifis-net repository?

@ruester I am open for suggestions. At this point I created description so you can replicate and test. If we agree that proposed approach covers all our needs then we can discuss location. I do prefer that our configuration is stored in the project repository as default. I assume that users who use RSD images (released RSD version) will mount their configuration in their production repo. If they do not mount any volumes our default configuration will be loaded (and RSD will work by default). 

In case of Helmholtz I am open for suggestions. Let me know what is your preference.

frontend/pages/index.tsx
ruester: so all future "home"-components need to have the same parameters.. so we do not use `{counts}` for now, but `HelmholtzHome` should be called with those props nevertheless

frontend/styles/themeDefault.js
```diff
@@ -1,98 +0,0 @@
...
-// SPDX-FileCopyrightText: 2022 dv4all
-//
-// SPDX-License-Identifier: Apache-2.0
-
-/**
- * Colors for both themes
- * Tailwind and MUI-5 based on MUI-5 definitions
- */
-
-const colors = {
```
ruester, line 13: I cannot find a replacement for those color definitions, where are they now? Same with `themeConfig.js`, how are the primary or secondary colors are defined now?

dmijatovic: @ruester the themes and colors, along with other settings are defined in 'settings.json' (frontend/public/data/settings.json). Maybe is handy to plan an Teams session this or next week? You can join our daily standup at any time. Do you have a link?

ruester: ah, I think it is done now via CSS and the tailwind configuration using CSS vars?

ruester: @dmijatovic yes, good idea, I will join you on Thursday at 10:00, since I have another meeting tomorrow :+1:

ruester:
On first thought, I believe this breaks the development of the frontend since a `yarn dev` does not mount the custom theme... But I am still onto getting our customization into the new structure..

frontend/styles/themeDefault.js
ruester: sorry, just found the `settings.json` which answers all my questions :wink: such a big PR :sweat_smile:

frontend/pages/_document.tsx
```diff
@@ -16,17 +16,25 @@ export default class MyDocument extends Document {
   readonly hotjarId = process.env.HOTJAR_ID
   readonly matomoUrl = process.env.MATOMO_URL
   readonly matomoId = process.env.MATOMO_ID
+  readonly fontFamily = process.env.RSD_FONT_HREF
```
ruester, line 19: this does not seem to be used in this file?

frontend/pages/_app.tsx
```diff
@@ -100,6 +111,8 @@ function RsdApp(props: MuiAppProps) {
   // console.log('session...', session)
   // console.log('settings...', settings)
   // console.log('matomo...', matomo)
+  console.log('rsdSettings...', rsdSettings)
```
ruester, line 114:
```suggestion
  // console.log('rsdSettings...', rsdSettings)
```

frontend/pages/index.tsx
```diff
@@ -5,271 +5,35 @@
...
 // fetching data server side
 // see documentation https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering
 export async function getServerSideProps() {
-  // extract params from page-query
+  // get counts for default rsd home page
   const counts = await getHomepageCounts()
-  // will be passed as props to page
-  // see params of SoftwareIndexPage function
+  // get host information for home page
+  const host = process.env.RSD_HOST ?? 'rsd'
```
ruester, line 32: this should be added/documented in the `.env.example` file too

ruester: also this variable is not used in this file... I think it got replaced by configuring it in the `data/settings.json` file?

ruester:
I think I got it working now, see: https://github.com/hifis-net/RSD-as-a-service/tree/custom-themes
but maybe some things are not quite working as before, I will check...

dmijatovic:
> I think I got it working now, see: https://github.com/hifis-net/RSD-as-a-service/tree/custom-themes but maybe some things are not quite working as before, I will check...

@ruester Thanks for you feedback. I am working on bug issues at the moment. I will check your branch this Friday or Monday next week and come back with my feedback.

ruester:
For reference what we just have talked about, here is an example for having the `yarn dev` running within a docker container:
- https://github.com/locationtech/geoperil/blob/master/docker/docker-compose-dev.yml#L69
- https://github.com/locationtech/geoperil/blob/master/frontend/Dockerfile.dev
- https://github.com/locationtech/geoperil/blob/master/frontend/start-dev.sh

I will test to adopt this same behaviour for the RSD frontend, so that we can solve working with a custom theme in development mode.

ruester:
With my last commit I was able to have hot reload for the frontend with our custom theme mounted via docker-compose (see https://github.com/hifis-net/RSD-as-a-service/commit/db77ab1447f4f09203e6c9235af16282308b1478) :+1: 
I now have some problems with the organisation logos, but I will debug this too :wink: 
I also had to change the way the Helmholtz home page component is called (some additional props).. which diverged in our fork from the way the RSD home page component is called :confused: 
https://github.com/hifis-net/RSD-as-a-service/commit/c160343ccd9a8a16f563a6754bca6309a9734154
Maybe we can also find a generic solution to add a custom home page component so that it does not need to be commited in the RSD repository? :thinking:

ruester:
I've now updated the "How to test" section for this PR because with the new approach the directories `frontend/{node_modules,.next}` will be created by the `root` user and could lead to errors if those directories have mixed ownership

nginx/nginx.conf
```diff
@@ -68,3 +68,60 @@ server {
 		proxy_pass http://backend/;
 	}
 }
+
+server {
```
ruester, line 72: I needed to serve the development mode of the frontend with nginx, so that all the other paths work too, because the frontend calls `localhost:3000/images` for example

ruester:
So all in all I got it working for our Helmholtz theme and home page. We may need to further implement generic approaches for the home page components, so that we end up with less merge conflicts over time, but this can be done later with some new issues/PRs.
Only some small remarks are left now from my side and if the new technique for the frontend development is also ok for you I would be happy to merge this PR :wink: :+1: 
Again many thanks for all those good ideas which you have implemented! It saves us some custom commits on top of our fork :+1:

nginx/nginx.conf
dmijatovic: @ruester Maybe this is not necessary? Next.js has proxy defined in next.config.js so it should proxy requests to http:localhost/image. Maybe enabling frontend-dev service on port 3000 and "re-defining" proxies is sufficient? I'll have look at this early next week. Important point is that this nginx.conf is used in the production image. If there is no other way than changing nginx.config we would need to have dev and prod nginx.conf files (and images).

dmijatovic:
@ruester I updated next config to properly proxy image, api and auth calls. I rebased to the latest main and needed to force push after that. This version should also work for you. Let me know if there are any other things we could improve.

We wait with merging of this PR until we have approval from @cmeessen, @ctwhome and @ewan-escience.

ruester:
> I updated next config to properly proxy image, api and auth calls.

Many thanks, it also works for our customization :+1:

ruester (approved):

ewan-escience (approved):
I tried it out without custom theme settings or volumes, and with the default theme everything still seems to work.

Makefile
```diff
@@ -7,14 +7,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # PHONY makes possible to call `make commands` from inside the Makefile
-.PHONY: dev-docs dev-frontend
+.PHONY: start install dev down dev-docs
```
ruester, line 10: is it good practice to list all targets here which are not a file to be generated, otherwise make checks the timestamp of a potential file (for example named "start") and will not execute those steps

> A phony target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request. There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.

ruester, line 10:
```suggestion
.PHONY: start install frontend data dev down dev-docs
```

ruester (approved):

frontend/package.json
```diff
@@ -14,9 +15,9 @@
     "backend:dev": "docker-compose up --scale test=0"
```
ruester, line 15: does this really work, since there is no `docker-compose.yml` within the `frontend` directory? or is this left over from some previous developments?

dmijatovic: Well spotted! I will remove it and implement your previous suggestion in a minute

cmeessen (approved):

dmijatovic:
@ctwhome and @jmaassen Can you have a look if the instructions in the README work on your machines and the text is clear?

README.md
```diff
@@ -26,41 +26,52 @@ SPDX-License-Identifier: CC-BY-4.0
...
-**Requirements:**
+You can run frontend in development mode, with hot-reloading, as docker service (frontend-dev). By default frontend-dev service is scaled to 0 (it will not be started automatically).
 
-- Docker
-- NodeJs ^16 or major
+```bash
+# frontend development as docker service at http://localhost:3000
+make frontend
+# OR
+docker-compose up --scale frontend-dev=1 --scale scrapers=0 -d
```
ctwhome, line 63: ❤️

ctwhome (changes requested):
This is a very big PR and complex. The work put in here has been fantastic. I want to add a remark on how we could think of a way to isolate completely the variations for different organizations like Helmholtz in this case, but also be able to support other organizations as well in the future. 

There is a bit of confusion on how things are growing organically. The major drawfall I see is that different variations of the application coexist within the base code rather than outside the main application. This can be a problem in the future if other organizations want to create their own themes/variations. Some individual components have subfolders with the name of rsd or helmholtz  (/components/home/).

We need to give it the last push to try to isolate variations from the base code.

deployment/README.md
```diff
@@ -29,6 +29,57 @@ RSD modules require a number of environment variables to work properly. The valu
 The default nginx.conf file is provided. The nginx image is based on nginx:1.21.6 with certbot already installed.
 To enable certbox certificate for your domain you will need to add your domains to nginx.conf file. docker-compose file expects nginx.conf file to be in the same folder.
 
+## Custom theme
```
ctwhome, line 32: this information should be in the documentation site

ctwhome (changes requested):

jmaassen (approved):
Go ahead and merge!

PR #510, opened 2022-09-07
Create a fake data generation script

### LINKED ISSUE #473: Generate fake data or import data from production
When #472 is merged, the data-migration script doesn't work anymore. This means that when developing locally or when testing out the dev vm, there is no data in those instances. We need a way to generate mock data and/or import data from the production instance. 

A problem with importing data from the production instance is that this needs to work between every version in development and whatever the version is at that moment in production. 

If someone knows a tool that can easily generate (realistic) data for a Postgres database, please let us know.

ewan-escience:
# Data generation

* A Node.js script is created that can generate fake data for the RSD. This data can locally be used for developing and testing purposes.

How to test:
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=1 --scale scrapers=0`
* Visit http://localhost and admire all the fake data
* Check if anything is missing

Closes #473

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2021 - 2022 Dusan Mijatovic (dv4all)
+# SPDX-FileCopyrightText: 2021 - 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+# SPDX-FileCopyrightText: 2021 - 2022 Netherlands eScience Center
 # SPDX-FileCopyrightText: 2021 - 2022 dv4all
-# SPDX-FileCopyrightText: 2021 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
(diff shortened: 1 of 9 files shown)


dmijatovic (approved):
Nice work! 

I have 2 suggestions:
* add `--scale scrapers=0` to docker-compose command to avoid starting scraper during data generation proces
*  set default number items to create to 50 or 100 in order to be able to test pagination, search and filtering on keywords

jmaassen (approved):
Look great, some of the titles are hilarious ;-)

Some possible improvements for the future:

- I get the same picture with all projects and organization. Could we generate different ones somehow?
- When browsing, the generated names of software and projects means I'm not always sure at which of the two I'm looking (as the text in itself is meaningless). For testing in may be an idea to always start with "software :" or "project :". And maybe also "organization : "?

PR #515, opened 2022-09-09
498 allow organisations multiple roles for projects

### LINKED ISSUE #498: mutual exclusion of funding and participating organizations
I am trying to add the eScience Center and another organization as funding organizations and participating organizations.

First, I add the organizations as funding organizations, I can save the project page and it works fine. But as soon as i add them to participating organizations, they disappear from the funding organizations part and appear only in participating organization part.

Since the eScience Center is both, we literally need that on all of our project pages.

Just in case, I am trying to edit the PADRE project page (https://research-software-directory.org/projects/the-petaflop-aartfaac-data-reduction-engine-padre) which was imported from the old RSD. But probably the same issue is with the newly created RSD project pages.

ewan-escience:
# Multiple roles for organisations of a project

* Organisations can now be both funding and participating for a project, while before they were mutually exclusive

How to test:
* Login and create a project
* Add a funding organisation to that project (in the "Information" tab) and save
* Add the *same* organisation to that project as a participating organisation (in the "Organisations tab")
* The funding organisation should still be there in the admin interface
* View the project and check that the organisation is shown both as funding and as participating
* Delete the organisation as a funding organisation, check it's still a participating organisation
* Add again and now delete it as participating organisation, do the same check
* On the organisations overview, check that the count is correct, i.e., check that the project isn't counted as 2
* Go to the organisation page and check that the project is only shown once
* Are there any other places that I missed where it could show up as a duplicate?

Note: when pinning or denying a project as organisation maintainer, this is done for all its roles. We still have to think if we want to change this later. There is also an edge case when a project is pinned or denied when it had only one role and then is later added with the other role.

Closes #498

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/012-inter-relation-tables.sql
+++ b/database/012-inter-relation-tables.sql
@@ -18,8 +18,8 @@ CREATE TYPE organisation_role AS ENUM (
 );
 
 CREATE TABLE software_for_software (
-	origin UUID references software (id),
-	relation UUID references software (id),
+	origin UUID REFERENCES software (id),
+	relation UUID REFERENCES software (id),
 	PRIMARY KEY (origin, relation)
 );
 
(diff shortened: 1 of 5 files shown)


dmijatovic (approved):
Nice job! I tested all mentioned scenarios with two organisations. They all worked properly.

PR #562, opened 2022-10-04
Improve ux serach boxes

### LINKED ISSUE #536: Ambiguous searching box when moving from organizations to software belonging to that organization
Based on feedback from @LourensVeen 

The searching box for an organization on the Organization page and the searching box for the software of an organization is placed in the same position, suggesting that both search boxes will do the same but they are completely different.

When going inside an organization both search boxes are placed in the same position:  
ambiguous

Since the search for software of projects are at the same level that the organization, it suggested that at least that search box will filter organizations, but instead it searches for software of projects. There is an overlap of context:

image

ctwhome:
Fixes #536

- This PR changes the placeholder from filtering projects, organisations and software
- Add search icon to the global searchbox
- Fix touch on mobile devices for the global search

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/organisation/project/index.tsx
+++ b/frontend/components/organisation/project/index.tsx
@@ -17,7 +17,7 @@ import {OrganisationComponentsProps} from '../OrganisationNavItems'
 
 export default function OrganisationProjects({organisation, isMaintainer}:OrganisationComponentsProps) {
   const {token} = useSession()
-  const {searchFor,page,rows,setCount} = usePaginationWithSearch('Search for projects')
+  const {searchFor,page,rows,setCount} = usePaginationWithSearch('Filter projects')
   const {loading, projects, count} = useOrganisationProjects({
     organisation: organisation.id,
     searchFor,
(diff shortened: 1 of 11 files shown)


dmijatovic (approved):
Nice work! Thanks.

frontend/components/organisation/project/index.tsx
```diff
@@ -17,7 +17,7 @@ import {OrganisationComponentsProps} from '../OrganisationNavItems'
 
 export default function OrganisationProjects({organisation, isMaintainer}:OrganisationComponentsProps) {
   const {token} = useSession()
-  const {searchFor,page,rows,setCount} = usePaginationWithSearch('Search for projects')
+  const {searchFor,page,rows,setCount} = usePaginationWithSearch('Filter projects')
```
dmijatovic, line 20: Maybe mention "this organisation"? Not sure if it will fit into available space, especially on the smaller screens.

ctwhome: We could leave as it is right now and see what the feedback is. I hope it is clear that it will filter in the current selected organization, but if not we could find a way

dmijatovic: Yes, sure.

PR #571, opened 2022-10-07
Feedback panel

### LINKED ISSUE #557: Add feedback button
Add a button Link to GitHub issues: 

image

image

ctwhome:
# Add a feedback panel at the app header
### UPDATE 3
Enable the setting in the `/frontend/public/data/settings.json` under the properties:
```
  "feedback": {
    "enabled": true,
    "url": "https://feedback.example.com",
    "issues_page_url": "https://github.com/research-software-directory/RSD-as-a-service/issues"
  }
```

### UPDATE 2
move popover to modal and UI modifications to the app header
Oct-14-2022 19-35-08

### UPDATE

Also, now the feedback button shows an email icon to "indicate" it wil be sending an email. It is not pretty obvious, hopefully, we can change this behavior in the future.
image

-------

Fixes #557  and improves UX on dropdown menus at the app header
image

* Creation of quick access -always present - menu feedback. The user can write the message and an email will be generated with the **text** and the **URL location** when the feedback is written
* The text area is automatically selected after opening the feedback panel
* The message is saved in case the user clicks away from the form accidentally to avoid losing their message.
* It is possible to send an email or open the GitHub issues from the more information message

image

image

How to test:

* go to any page of the application and click on the menu feedback to open the form. Write a message and sen

PR Checklist:

*   [*] Link to a GitHub issue
*   [*] Update documentation
*   [ ] Tests


ctwhome:
the base URL is fixed to [https://github.com/research-software-directory/RSD-as-a-service/issues](https://github.com/research-software-directory/RSD-as-a-service/issues) @dmijatovic is there a way to get this from .env?

cmeessen:
That's a great idea to lower the barrier of sending feedback. I'd be happy if we could configure whether the feedback button should be shown at all, and which email address should be used.

Also, how about to directly store the feedback in the database? I can't really estimate how many problems we could get from bots or malicious attacks, but this would remove the barrier of having to send a mail. Some ideas to stop abuse of the function: minimum/max character count, limit amount of requests per minute for one user, limit total amount of requests per minute on server side. It would also come with some effort to make a frontend page for administrators to manage the reviews.

ctwhome:
thanks @cmeessen those are great suggestions. For now we will just go with the email but keeping in mind the admin panel for reviewing feedback, for example, and also adding a mail server to not have to open the users email to send the feedback. This should be only a remporal preliminary solution.

ctwhome:
@dmijatovic it seems to be like the env variable is not being catch up by the production version, on the dev (http://localhost:3000/) it works. I double check if the envs were exposed in the docker compose but I do'n tsee anything unusual. 
woudl you mind to double check it to see where could be the problem?

.env.example
```diff
@@ -180,3 +180,8 @@ CROSSREF_CONTACT_EMAIL=
 # URL (should end with a trailing slash) and ID for Matomo Tracking Code
 MATOMO_URL=
 MATOMO_ID=
+
+
+# consumed by: frontend
+# Show feedback panel in at the top header. Leave it empty to don't show it to hide the feedback panel.
+NEXT_PUBLIC_FEEDBACK_URL=      # rsd@esciencecenter.nl
```
dmijatovic, line 187: Please remove env variable. We use settings.json, property "feedback_url"

docker-compose.yml
```diff
@@ -112,6 +112,7 @@ services:
       - NEXT_PUBLIC_HELMHOLTZAAI_WELL_KNOWN_URL
       - NEXT_PUBLIC_HELMHOLTZAAI_SCOPES
       - NEXT_PUBLIC_HELMHOLTZAAI_RESPONSE_MODE
+      - NEXT_PUBLIC_FEEDBACK_URL
```
dmijatovic, line 115: revemove "NEXT_PUBLIC_FEEDBACK_URL"

ewan-escience (changes requested):
-  When commenting out `NEXT_PUBLIC_FEEDBACK_URL`, it still shows up. I heard this is not used anymore, because a build time variable from `frontend/config/defaultSettings.json` is used instead, so it can be removed from the `.env.example`.
- When sending the mail with Thunderbird, the message body is empty.
- The link to the GitHub issues doesn't seem to be configurable?

dmijatovic (approved):

ewan-escience (approved):
Thunderbird still doesn't work, but that seems to be unfixable.

PR #573, opened 2022-10-10
556 orcid login

### LINKED ISSUE #556: ORCID login
To facilitate people from universities outside of the Netherlands, we should provide an OpenID provider different from SURFconext. We decided to try out ORCID. 

If we do this, in principle anybody could login, so we should have a whitelist of ORCID's (or some other approval system) to limit who can login. Options are: 
- A whitelist of ORCID's in the database, you can only login if you're in that whitelist
- You can always login but you cannot add something unless you're on the whitelist, this would require a different database role (or you don't get a role, effectively you get the same right as an unauthenticated visitor)

Ideas on how to get on the whitelist: 
- Send us an email with your ORCID and we enter it into the whitelist
- Send a mail and we send an invite link back
- With the second option of the first list, after logging in, click on a request button, either the RSD or an organisation approves
- An existing user sends an invite link
- You are already logged in with SURFconext, you can then login with ORCID

ewan-escience:
# Allow users to login with ORCID

* Users can login with ORCID if they are on the whitelist in the database
* Add a page where admins can add or delete whitelisted ORCIDs
* Fix bug: if you add something to RSD_AUTH_PROVIDERS that we don't support, the sign in button doesn't show up

How to test:
* Copy the new values from `.env.example` to `.env` and fill in the secret (contact us or create an integration on the [ORCID sandbox environment](https://sandbox.orcid.org/))
* Add `ORCID` to the `RSD_AUTH_PROVIDERS` env variable
* Also add something nonsensical to the `RSD_AUTH_PROVIDERS` env variable
* Login with (a sandbox) ORCID, this should fail as you're not whitelisted
* Login as an admin, add your ORCID to the whitelist (and play around on that page)
* Login with your whitelisted ORCID, this time it should work

Closes #556

PR Checklist:

*   [ ] Update documentation
*   [x] Tests


cmeessen (approved):
My suggestion is to add a comment about allowed values in `RSD_AUTH_PROVIDERS` (I can't make a code suggestion, so I'll just write it here, last line):

```
# consumed by services: authentication, frontend (api/fe)
# provide a list of supported OpenID auth providers
# the values should be separated by semicolon (;)
# Allowed values are: SURFCONEXT, HELMHOLTZAAI, ORCID or LOCAL
```

Otherwise: nice work, works as described :+1:

jmaassen (approved):
Works as expected! This should be enough to get started with ORCID.

PR #600, opened 2022-10-21
597 sanitise keywords and improve checks

### LINKED ISSUE #596: Duplicate/ghost keywords from `Import Keywords`
**OS**: Windows 10. **Browser**: Firefox 105.0.3

When editing a software page, I added keywords manually. Later I clicked on "Import Keywords" to import from github/zenodo, and many duplicate keywords showed up:

<details>
<summary>Before:</summary>
<br>

image

</details>

<details>
<summary>After:</summary>
<br>

image

</details>

The first time I did this, the keywords did not disappear, but instead duplicated when trying to close them. I was able to recreate this bug this once on a call with @ctwhome, but was not able to record it.

### LINKED ISSUE #597: Duplicate `python` keyword suggestion.
When adding keywords, and typing "pyth...", suggestions come up. However, Python is listed twice:
image

dmijatovic:
# Additional validations on the keyword entries and deduplication

Closes #596 
Closes #597

* We do not allow leading & trailing spaces or multiple spaces between the words. The database will reject these entries
* On input side we improved search (without leading/trailing spaces) and deduplication of imported keywords
* The keywords are trimmed before send to api
* On the presentation side we use capitalized value in the "chips", this does not impact accronims, for example GPU is still shown in capitals

How to test:
* `make start` to reset and rebuild
* login as rsd_admin or create new software or project
* navigate to software and try to add Python and python with some spaces before or after the keyword. All the results should produce one python entry which in "chip" will be shown with capital P.

## Example of capitalized keyword presentation on edit page
image

## Example of capitalized keyword on software page
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/006-create-keyword.sql
+++ b/database/006-create-keyword.sql
@@ -5,7 +5,7 @@
 
 CREATE TABLE keyword (
 	id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
-	value CITEXT UNIQUE
+	value CITEXT UNIQUE CHECK (value ~ '^\S+( \S+)*$')
 );
 
 CREATE FUNCTION sanitise_insert_keyword() RETURNS TRIGGER LANGUAGE plpgsql AS
--- a/database/020-row-level-security.sql
+++ b/database/020-row-level-security.sql
@@ -241,6 +241,9 @@ CREATE POLICY anyone_can_read ON keyword FOR SELECT TO web_anon, rsd_user
 CREATE POLICY maintainer_can_insert ON keyword FOR INSERT TO rsd_user
 	WITH CHECK (TRUE);
 
+CREATE POLICY maintainer_can_delete ON keyword FOR DELETE TO rsd_user
+	USING (TRUE);
+
 CREATE POLICY admin_all_rights ON keyword TO rsd_admin
 	USING (TRUE)
 	WITH CHECK (TRUE);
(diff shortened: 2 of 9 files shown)


ewan-escience (changes requested):
Please don't change the casing when storing keywords into the database, this can break acronyms. I'd also prefer to show them as they are stored. For the rest, it works well.

dmijatovic:
> Please don't change the casing when storing keywords into the database, this can break acronyms. I'd also prefer to show them as they are stored. For the rest, it works well.

@ewan-escience I changed save to original casing. Concerning the capitalize, I just checked, it ignores keywords like GPU and with improvement of saving I think we have optimal approach.

ewan-escience (approved):

jmaassen:
It's a bit weird to get an error when I type a keyword that has 1 hit in the database already. For example, if I type "GPU", then there is a hit in the database, and I'm forced to click the "GPU" found in the database to add it.

To me, it would make more sense to automatically add the database "GPU", since it's a perfect match?

jmaassen:
Found a small bug (I think): 

- add a new keyword (like "hello")
- try to add it again, it will say it's already in the list (which is correct)
- delete it from the list
- add it again. It won't show up as being in the database already, but will give an error when trying to add it. 
- after a refresh, it does show up as being in the database

dmijatovic:
> Found a small bug (I think):
> 
> * add a new keyword (like "hello")
> * try to add it again, it will say it's already in the list (which is correct)
> * delete it from the list
> * add it again. It won't show up as being in the database already, but will give an error when trying to add it.
> * after a refresh, it does show up as being in the database

@jmaassen Well spotted! I further improved the validations. Please confirm improvements

jmaassen (approved):
The bug is now gone, and auto selecting the keyword when it matches also work fine!

PR #624, opened 2022-11-01
622: show only specific links in the organisation info card

### LINKED ISSUE #622: organization page - org url duplicate
For the escience center, the organization page has recently been improved (thanks!).
Howerver, we now show not-so-relevant links for NLeSC:
```
http://www.esciencecenter.nl/
https://www.esciencecenter.nl/
https://ror.org/00rbjv475
```
Can we remove duplicate links? So don't show URLs that only differ in the protocol (http vs https)?

2) The styling of the organization type is too agressive: the 'nonprofit' badge is roughly half the size of the NLeSC logo; and the green is quite ugly next to our NLeSC blue. Less contrasting colors (black on gray?) and a smaller size for the org type would be useful.

Atlernatively, both the 'Type' label, the 'nonprofit' badge, and the 'https://ror.org/XXXX' link could be a single badge/button/...to the ror page.
And the NLeSC logo could be a link to www.esciencecenter.nl , such that the whole 'links' could go.
Or are there plans to use the 'Links' section for more informative urls?

dmijatovic:
# Show only specific links in the organisation card

Closes #622

* The links in the organisation card are selectively applied because the links received from ROR api are in some cases outdated (use http instead of https). We only show: first website link, link to ROR info and, if provided, wikipedia link.
* Instead of showing links we show an icon and the label: Website, ROR info and Wikipedia
* New logic selects the links in the following way: 
    * website link: if the website property is present in the organisation in RSD we use that link, otherwise we use first entry from the links array returned by ROR api. If both are not available we do not show website link entry.
   * ROR info: if ror_id is provided by ROR api we show this link as "ROR info" (id is always present)
   * Wikipedia: if wikipedia link is provided by ROR api we show this entry under "Wikipedia"
* Organisation type label is made (slightly) smaller by reducing the font size
* The color of organisation type label remains unchanged (it is primary color). It can be made neutral (which has less flexibility when changing the theme colors). See second printscreen and let me know what you prefer.
* **Unrelated change**: commit graph message changed to "We cannot display the activity graph, because we do not know the repository url."

How to test:
* `make start` to restart everything, when you done use `docker-compose down`
* login as rsd_admin to be able to edit existing project/software, otherwise create new software/project
* add organisation, for example eScience Center (with logo), VU or other
* verify the organisation card entries
* any suggestions are welcome!

## Example VU organisation card
image

## Example organisation type with neutral color
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ewan-escience (approved):
Nice work! I prefer the green label by the way.

cmeessen (approved):
Works and looks great :+1: 

My preference for the displayed webpage would be, to only use the domain name without path, because in ROR there are sometimes really strange links. Here is an example: The [ROR entry for the German Aerospace Center](https://ror.org/04bwf3e34) links to http://www.dlr.de/dlr/en/desktopdefault.aspx/tabid-10002/. I think the risk that such specific links will stop working/be outdated at some point is quite high.

dmijatovic:
> My preference for the displayed webpage would be, to only use the domain name without path, because in ROR there are sometimes really strange links. 

@cmeessen Thanks for the feedback. Concerning the "strange" url I would not modify ROR info. If organisation wants to use different website link it can adjust ROR entry or organisation admin can define it in RSD (we use RSD link instead of ROR link when provided). In addition, using only domain will not work for some ROR organisations, see Microsoft example below.

image

cmeessen:
I just pushed a minor change to replace the message for missing URL in the commit graph replacement. I had the impression that 'the repository url is missing' sounds a bit rude, and changed it to "We cannot display the activity graph, because we do not know the repository url."

jmaassen (approved):
Looks great!

PR #632, opened 2022-11-04
620 retry commit scraper on 202

### LINKED ISSUE #620: Commit scraper should retry on HTTP status code 202
When scraping commit data from GitHub, sometimes we receive a HTTP 202 response. This means that GitHub will start to generate the data but will not send it in the reply. Instead, we should obtain the data later. Our scraper should make one retry to get this data in the same scraping round. We should look if there is a header telling us when to retry, otherwise we should wait for a small time period.

This should hopefully fix the problem where for new software entries, the graph shows that the data could not be scraped, even if an open GitHub link is provided.

ewan-escience:
# Retry commit scraper on HTTP status code 202 from GitHub

* If we get status 202 from GitHub when scraping commits, wait 3 seconds and try again
* If we then still get 202, we treat it as if we could not scrape it and reuse the old data

How to test:
* This is hard to test, since you need a 202 response from GitHub and we don't print to console when this happens, I manually tested this (also with prints)
* Check that it builds at least
	* `docker-compose build auth && docker-compose up`
* Make sure to have some repos in the database and manually run the commit scraper a few times to see no unexpected errors occur:
	* `docker-compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.git.MainCommits`

Closes #620

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,7 +130,7 @@ services:
 
   scrapers:
     build: ./scrapers
-    image: rsd/scrapers:1.2.0
+    image: rsd/scrapers:1.2.1
     environment:
       # it uses values from .env file
       - POSTGREST_URL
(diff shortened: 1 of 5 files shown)


dmijatovic (approved):
Works well as far as I can validate

PR #637, opened 2022-11-07
589 token url

### LINKED ISSUE #589: Remove *_TOKEN_URL from .env
In our `.env` files (see `.env.example`), we use *_TOKEN_URL environment variables. The values of these variables contain the URL of where to send the OpenID code to when a user logs in. However, every OpenID provider also provides a "well known" discovery endpoint, which also contains this token URL. We also have the URLs of these discovery endpoints in our `.env` files (see the `*_WELL_KNOWN_URL` env variables).

It would be better to get rid of the token env variables and use the well known env variables only, both to prevent configuration errors (e.g. a mismatching well known and token endpoint) and to dynamically react to changes of these endpoints.

We could either query the well known endpoint for the token endpoint on every login request or store this URL in memory and periodically (say every 15 minutes) update it in the background. In the first case, the performance impact should be tested first.

ewan-escience:
# Remove token endpoints from config

* Remove the `*_TOKEN_URL` env variables and adapt the auth module so that the token URL is queried from the well known URL on every request
* Give the reason when you login with e.g. GitHub through Helmholtz AAI and `HELMHOLTZAAI_ALLOW_EXTERNAL_USERS=false`

How to test:
* Remove `*_TOKEN_URL` values from your `.env` (see `env.example`)
* `docker-compose build auth && docker-compose up`
* Set `HELMHOLTZAAI_ALLOW_EXTERNAL_USERS=true`
* Login with all three providers, this should work
* Set `HELMHOLTZAAI_ALLOW_EXTERNAL_USERS=false`
* Login with Helmholtz AAI through GitHub or ORCID, you should see in red text `You are not allowed to login`

### Discussion point
Maybe we should cache the token URL instead of querying it on every login. On my tests, ORCID was slow, taking 200 -- 900 ms. Helmholtz and SURF were below 100 ms.

Closes #589

PR Checklist:

*   [ ] Update documentation
*   [x] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -82,8 +82,6 @@ NEXT_PUBLIC_SURFCONEXT_WELL_KNOWN_URL=https://connect.test.surfconext.nl/.well-k
 NEXT_PUBLIC_SURFCONEXT_SCOPES=openid
 # consumed by: frontend/utils/loginHelpers
 NEXT_PUBLIC_SURFCONEXT_RESPONSE_MODE=form_post
-# consumed by services: authentication
-AUTH_SURFCONEXT_TOKEN_URL=https://connect.test.surfconext.nl/oidc/token
 
 # Helmholtz AAI
 # consumed by: authentication, frontend/utils/loginHelpers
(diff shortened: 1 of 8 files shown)


dmijatovic (approved):
Nice work!

cmeessen (approved):
Works like a charm, the new error messages are really helpful :+1:

PR #648, opened 2022-11-10
385 filter research domain

### LINKED ISSUE #385: Search for software and projects needs sorting/filtering
When searching for either software or projects, we need more than only a single text entry box.
Also, what exactly is being searched? It seems like project or software title + description? But not the rest of the meta data?

It would be useful to be able to sort and filter on at least:

For projects:
* running / finished (might not be needed if 'finished' is sorted to the end)
* domain

For software:
* expertise / technology used
* programming language?

Both:
* organizations
* contributors

After discussion with Dusan:
Add a 'Research Domain Page', accessible via a url. $RSD/researchdomain/TAG
where TAG is sort-of drill down;
* PE, would cover all PE1, PE1_1 etc
* PE1 would cover only PE1 and PE1_*
* etc.
note that if project have fi. the PE1_1 label, they need not have PE.

dmijatovic:
# Add filtering on research domain to projects page

Closes #385 

*  User can filter for projects with a specific research domain from the project (item) page by clicking on the research domain tag. See example 1 screenshot
* User can filter projects for a specific research domain or combination of keywords and research domains on the project overview page. When filter is active the search input is applied within the filter selection. When filter is not applied the search term is also applied to keyword and research domain values of the project.
* Scaling image label is changed to: "Fit image into the area (recommended when using logos)" 
* Global search improved to include title+subtitle+keywords+research domain wild card search in software and projects while for organisation we search in name and the website.

How to test:
* `make start` to rebuild everything, when done use `docker-compose down`
* navigate to project page with research domain tags, you should be able to click on research domain which will open project overview page with the filter on selected research domain
* on project overview page you can select multiple filters and clear those
* try searching by keyword or research domain values, this should also work when filter is not applied.
* similar results should be achieved in global search

## Example 1 - research domain filter on project page
image

## Example 2 - research domain filter on project overview page
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen:
Small comment about the info box in project search. It now says: 

```
Filter is not active
Select keyword or research domain from the lists or type in to search.
```

would be better to use: 

```
No filter active 
Select a keyword or research domain from the list, or type to search.
```

jmaassen:
Another textual comment. When a filter is selected, the search bar shows "Find within selection". It may be better to use "Search within selection"

jmaassen (approved):
Nice improved. Some textual changes suggested

PR #652, opened 2022-11-14
docs: adds info about logging in with admin rights

cmeessen:
# Instructions how to login with admin rights

Due to an unforeseen downtime of the dev instance of Helmholtz AAI, I was forced to use other login methods. I used the opportunity and updated the docs accordingly.

* adds a section about 'Logging in with admin rights'

How to test:

```
cd documentation
yarn install
yarn dev
```

or `make dev-docs`

Open http://localhost:3030 and navigate to "Getting Started" -> "Running locally" -> "Logging in with admin rights"

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/documentation/docs/getting-started.md
+++ b/documentation/docs/getting-started.md
@@ -1,5 +1,7 @@
 <!--
+SPDX-FileCopyrightText: 2022 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 SPDX-FileCopyrightText: 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+SPDX-FileCopyrightText: 2022 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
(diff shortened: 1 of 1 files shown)


dmijatovic (approved):
Nice addtion, thanks!

ewan-escience (approved):
Nice addition, maybe you could change the `danger` part to use actual Markdown so that it renders differently.

cmeessen:
> Nice addition, maybe you could change the `danger` part to use actual Markdown so that it renders differently.

I optimised it for the rendered documentation which now looks like this:

image

Would you prefer that the documentation is markdown only?

ewan-escience:
I only tried rendering it in my IDE (IntelliJ IDEA) and on https://markdownlivepreview.com/, and in both cases it doesn't render as nicely as in your screenshot. Where is that syntax from then? And if it shows like that in the documentation, it's fine.

cmeessen:
These are Vuepress [custom containers](https://vuepress.vuejs.org/guide/markdown.html#custom-containers)

PR #658, opened 2022-11-15
New setup for saving images

dmijatovic:
# Refactor image storage and reference

**This issue is work in progress.**

Closes #631 #635 #636 #656

Changes proposed in this pull request:
* Upgrade to Next v13
* Images are stored in one table. The id is SHA-1 of the image data. This addresses the caching in the most optimal way for an SPA (reloading of page is not required when user changes image).
* Search for contributors and team members extended with avatar image. Note! This feature is only supported for RSD persons as ORCID api does not provide avatar images.
* Search for contributors and team members in RSD includes both contributors and team members
* Data generation script is updated to support new image approach
 
How to test:
- `make start` to rebuild and generate some random data
- validate image uploads on all places:
- for software: contributors avatar upload, organisation logo upload
- for projects: project image, team member avatar upload, organisation logo upload
- for organisations: organisation logo upload, research unit organisation logo uploadPR Checklist
- add a person as software contributor, then go to project to team members and look for this person. You should be able to find this person in RSD
- when person with avatar is removed from all references the avatar image should also be removed from image table
- when project image is removed from all references the image should also be removed from image table

*   [ ] Increase version numbers in `docker-compose.yml`
*   [ ] Link to a GitHub issue
*   [ ] Update documentation
*   [ ] Tests

PR #676, opened 2022-11-28
fix: date in future test

cmeessen:
# Fixes a test that fails  because date is not in the future anymore

The test `returns 0 if date in future` in `frontend/utils/dateFn.test.ts` was failing, because the date it was testing against is 2022-11-25, which is not in the future anymore.

* use `2122-11-25` as date

How to test:

* make sure that all images that were compiled before 2022-11-25 are removed (e.g. by using `docker system prune`, `docker rmi` was not sufficient for me because of layer caching)
* `docker compose build frontend`

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


cmeessen:
Since the fix in #658 is much more suitable, I will close this PR so we do not add unnecessary noise to the commit history.

PR #689, opened 2022-11-30
Do not set scrapedAt if rate limit is exceeded

### LINKED ISSUE #686: Change no-commit-graph message if API rate limit is reached
If the API rate limit is reached when a repository is scraped for the first time, the commit graph says "We cannot display this graph because we cannot read the commit history."

In this case, the message should be that the commit graph was not scraped yet.

cmeessen:
Fixes #686 

* if the response code indicates that the API rate limit is exceeded (403 for github, 429 for GitLab), do not update the entry so that it will still be listed as a repo that has not been scraped yet

How to test:

* make sure that `API_CREDENTIALS_GITHUB` in `.env` is unset
* `docker compose build scrapers && docker compose up`
* reach the rate limit by running
  ```
  for i in {1..60}; do curl -I https://api.github.com/users/octocat ; done
  ```
* If the rate limit is reached, create a new software with url `https://github.com/research-software-directory/RSD-as-a-service`
* Run the scrapers `docker compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.git.MainCommits`
* Open the corresponding software page, the graph should show the "we did not scrape the history yet" message

The check is also integrated for GitLab repositories, however, I could not manage to reach a limit.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/scrapers/src/main/java/nl/esciencecenter/rsd/scraper/RsdRateLimitException.java
+++ b/scrapers/src/main/java/nl/esciencecenter/rsd/scraper/RsdRateLimitException.java
@@ -1,0 +1,12 @@
+// SPDX-FileCopyrightText: 2022 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+// SPDX-FileCopyrightText: 2022 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package nl.esciencecenter.rsd.scraper;
+
+public class RsdRateLimitException extends RuntimeException {
+	public RsdRateLimitException(String message) {
+		super(message);
+	}
+}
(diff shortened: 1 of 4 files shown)


ewan-escience:
We indeed get [a 403 status code](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limit-http-headers) from GitHub when we hit a rate limit, which is unfortunate. It would be better in that case to check the `x-ratelimit-remaining` header to see if this is indeed the case, but that requires a bigger refactoring (which I plan to do soonish 😀).

cmeessen:
> The code looks ok, but can you rebase this branch on main, so we can be sure these changes are fully compatible?

Pardon for that, I rebased it :)

ewan-escience (approved):

PR #690, opened 2022-11-30
Minor UI fixes

dmijatovic:
# Two minor UI fixes 

*  Project card to show image on small screens. This feature was lost during the last refactoring, we reverted to the version from previous release (v1.11.1)
*  Project card layout on mobile is further improved, the card orientation is changed to portrait for mobile while on larger screens remains to be landscape
*  Contact card layout is improved concerning the margins and the role of contact person is added
*  Software card and organisation card have similar approach for small screens (mobile, laptop portrait, laptop landcape)
*  Page title is not fixed when device height or width is lower than 640px

How to test:
* `make start` to rebuild everything and create example entries
* navigate to projects overview, the images should be visible even when you switch to mobile view
* contact person card will show the role if the role is specified

## Example project card in portrait for mobile
image

## Example project card in landscape for tablet
image

## Example project card for desktop
image

## Example organisation card in portrait for mobile
image

## Example organisation card for tablet
image

## Example organisation card in portrait for desktop
image

## Example contact person card
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen:
When moving to a portrait layout, the images which are not "shown fully" are often cropped in such a way that only a very small part of the image is shown. It works better for the fully shown images. 

I have similar issue with some the project pages themselves. Maybe we could consider cropping/scaling differently?

jmaassen:
For the projects it wraps the title to two lines when needed. For software and organization it crop them however. Wouldn't wrapping also be better there?

cmeessen (approved):
Works for me :+1:

I have one suggestion that could be addressed in a new issue: in landscape mode on mobile devices, the sticky header still shows when scrolling down, taking up more than 1/3 of the screen space:

image

In portrait mode, this header disappears completely. From my perspective it would make sense to also let it disappear in landscape mode.

jmaassen (approved):
Works well now the text is wrapped.

ewan-escience (approved):

PR #712, opened 2023-01-27
705 tos agreement

### LINKED ISSUE #705: Add option to ask for agreement to Terms of Service
Our data protection lawyer strongly advised us to actively ask for the agreement to the terms of service by users that want to contribute content. We would like to integrate this functionality by asking for agreement, e.g. a check box or an "Agree" button, directly after the first time a user logged into the RSD.

@jmaassen would you be interested in such a feature as well?

cmeessen:
# Require users to agree to Terms and Data Privacy

Fixes #705

Changes proposed in this pull reques:

* require users to agree to Terms of Service and state that they have read the data privacy declaration
* users cannot perform any action until they agree to all the documents
* if the user did not yet agree, a modal will pop up wherever something could be added or edited
* renamed 
* the user agreement can additionally be configured in the user profile
* a "Profile settings" was added to the user menu

How to test:

* clean volumes, rebuild database/frontend and start
  ```
  docker-compose down --volumes
  docker-compose build database frontend
  docker-compose up --scale scrapers=0
  ```
* login as a new user, e.g. as a local user
* try to add a new software, this modal will popup 
  modal
* click on Cancel -> you will be returned to page you came from
* try to add software again
* agree to the Terms of Service and the Privacy Statement, the OK button will turn green, click ok
* add a new software
* open the user Menu -> Profile Settings
* the switches below "User agreements" should be activated
* deactivate the switches
* Go to "My Software" and try to edit the software -> modal shows
* the same happens when trying to edit organisations where the user is maintainer of or when creating/editing projects

Things that should/could be improved

~~* the check whether a user agreed to all the documents is performed in the database. Maybe the NLESC doesn't require the agreement to three documents (e.g. if you will only have Terms of Service), so a configuration about which documents are required could be useful. Since checks are done in the DB, it might be worth to have some configuration options in the DB as well? (@ewan-escience ?)~~
~~* If the user logs in the first time, they should be forwarded directly to a separate page where they will be asked to document their agreement~~

To do

* [x] add check for DELETE
* [x] Rename Data Protection Declaration to Privacy Statement

Many thanks to @dmijatovic for fixing the issue with the forms :)

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ctwhome:
I suggest changing the way errors are displayed. Instead of an error message, an informational box or modal should be used to notify the user that the terms must be accepted before any changes can be made. This way, the user's modifications won't be lost if they need to leave the page to agree to the terms.

dmijatovic:
Maybe we can have modal with notification and the options so user can accept terms directly from the modal (no need to go to profile page)? The profile page form is still needed because user should be able to see and change these values at any time.

I believe data generation script and end-to-end tests are also impacted by this change.

ctwhome:
This is a modal suggestion to show after someone login: 
image

jmaassen (approved):
Works well. Two small points:

- I'm not sure the need the comma in "agree to our Terms of Service, and "
- It's a bit weird that the RSD administrator also needs to accept the TOS/PS.  Not a problem though, so it's fine to keep it like this.

ewan-escience (approved):

PR #722, opened 2023-02-07
651 organisation names

### LINKED ISSUE #651: Organisational units with same names are not distinguishable upon linking
When adding organisational units to two organisations, each with the same name (e.g. "Remote sensing") they are not distinguishable when trying to add them to a software:

image

ewan-escience:
# Show parent organisation names on hover

* When searching for an organisation to add to software or projects, on hover you will see the names of all the parents of the organisation
* Fix bug: when an organisation doesn't have a ROR ID, don't try to deduplicate on it

How to test:.
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=0`.
* Login as an admin and create a software page and make it published. Add two organisations to it. 
* Add to both organisations a research unit with the same name.
* On the software page, search for that name. You should see the two units, but when you hover over them, you will see the names of the parent organisations.
* Do the same for a project.

Closes #651

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


cmeessen (approved):
Works as described :+1:

jmaassen:
Works well. 

I do wonder if this will work with large numbers of organizations with clashing names (for example, most universities have a "law faculty"). But for now it is fine.

frontend/types/AutocompleteOptions.ts
```diff
@@ -1,12 +1,15 @@
...
+// SPDX-FileCopyrightText: 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2023 Netherlands eScience Center
 //
 // SPDX-License-Identifier: Apache-2.0
 
 export type AutocompleteOption<T> = {
   key: string
   label: string
   data: T
+  hover?: string
```
dmijatovic, line 12: AutocompleteOption is generic type. Can we put hover to be one of the properties delivered in the data prop and extract it from there where it's needed?

dmijatovic (commented):
I have one improvement suggestion

dmijatovic (approved):
I think this is good starting point. I am not sure about adding hover property to generic option type.

PR #726, opened 2023-02-10
Rpc wordpress improvement

ewan-escience:
# Show child organisation names in child list RPC

* The RPC called `list_child_organisations` also returns the organisation names
* Allow superusers to execute DELETE statements

How to test:.
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=0`.
* Run the following query and check that the results from the last two `SELECT` statements are correct:
```SQL
INSERT INTO organisation(slug, name, parent) VALUES
('parent', 'Parent organisation', NULL);

INSERT INTO organisation(slug, name, parent) VALUES
('child1', 'Child 1 organisation', (SELECT id FROM organisation WHERE slug = 'parent'));

INSERT INTO organisation(slug, name, parent) VALUES
('child1-1', 'Child 1.1 organisation', (SELECT id FROM organisation WHERE slug = 'child1'));

INSERT INTO organisation(slug, name, parent) VALUES
('child2', 'Child 2 organisation', (SELECT id FROM organisation WHERE slug = 'parent'));

SELECT * FROM list_child_organisations((SELECT id FROM organisation WHERE slug = 'parent'));

-- compare the data above with the data below
SELECT id, name FROM organisation;
```

Copy the UUID of the parent and fill it in the URL below, check that the results are correct:
http://localhost/api/v1/rpc/list_child_organisations?parent_id=uuid-of-parent

Run the following statement, it should run without errors:
```SQL
DELETE FROM organisation;
```

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/012-create-organisation-table.sql
+++ b/database/012-create-organisation-table.sql
@@ -96,21 +96,23 @@ $$;
 CREATE TRIGGER sanitise_update_organisation BEFORE UPDATE ON organisation FOR EACH ROW EXECUTE PROCEDURE sanitise_update_organisation();
 
 -- including the parent itself
-CREATE FUNCTION list_child_organisations(parent_id UUID) RETURNS TABLE (organisation_id UUID) STABLE LANGUAGE plpgsql AS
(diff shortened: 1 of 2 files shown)


dmijatovic (approved):
Nice work!

PR #750, opened 2023-02-27
fix: change RPCs to use SQL instead of PL/pgSQL to increase performance

ewan-escience:
# Improve RPC performance

* Change language of the function `organisations_overview`, and all its dependencies, from `PL/pgSQL` to `SQL`, so they can be inlined. This increases performance when only a subset of the columns or rows is requested.
* Adapt the frontend organisation overview page so that it only requests the columns it needs, in particular, don't request `release_cnt`, since this is an expensive operation.

How to test:
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=1`.
* Check http://localhost/organisations, and check various organisation pages
* Navigation should be quicker than on https://research-software.dev/organisations, which you can also see on the network tab of your browser's devtools.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/009-create-mention-table.sql
+++ b/database/009-create-mention-table.sql
@@ -80,7 +80,7 @@ CREATE TABLE mention_for_software (
 );
 
 
-CREATE OR REPLACE FUNCTION search_mentions_for_software(software_id UUID, search_text VARCHAR) RETURNS SETOF mention STABLE LANGUAGE plpgsql AS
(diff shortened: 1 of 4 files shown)


cmeessen (approved):

dmijatovic:
Nice work! The loadtests I performed locally show significant improvement! Loadtime of organisation overview is reduced by more than 50%.

image

dmijatovic (approved):

PR #790, opened 2023-03-21
feat: add page that gives overview of project quality

### LINKED ISSUE #764: Feedback from Jisk: overview of quality of pages one maintains
I got some feedback from @jiskattema after my demo at the SSH section. 

As a PM, he tries to keep the project pages nicely up-to-date, but given that fact that he is now maintaining 80+ of these, it is difficult to have an overview of what needs to be done where. 

He asked if it would be possible to have some quality indicator for the pages. Something like "your page is 65% complete", and then produce a view that can show this metric for a list of pages he maintains. 

This sounds quite similar to what a tool like fairtally is doing: 

Screenshot 2023-03-02 at 17-05-50 Fairtally results

We could think about a similar approach that allows a user to view a list of the pages they maintain, combined with one or more quality metrics?

ewan-escience:
# Add project quality overview page

* Add a page that lists all projects you maintain and its completeness
* On that page there is a big table which you can sort per column, including a score column that is the amount of columns with a non-zero or non-false value
* As an admin, you can either see projects you maintain or all projects
* Adapt the data-generation script so that some data is not always present

How to test:
* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=1`.
* Login as admin and go to http://localhost/user/project-quality
* Since you don't maintain any projects, you should see nothing, so click the switch
* Play around with the table, sort some columns and check on some pages that values in the table correspond to the page
* Become a maintainer of some pages and check that when the switch is turned off, you do see those projects in the table
* Login as a regular user and visit the same page
* You don't get to see the switch, so create some project pages and check again
* Click on you avatar and go to "My {anything}", you should see a link to the project quality page overview

Note, this is a rough implementation, but it should work and be good enough to be usable and to gather feedback on it.

Closes #764

## With added full screen button (bottom-right) 

image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen (approved):
Very nice!  Could use a bit of UX sprinkled on it though ;-)

At the moment I needed to scroll horizontally. The layout seems to use quite a lot of white space however, and it seems that improving the layout may make it fit easily?

Lets improve this later though. It may be good to get feeback from Jisk first?

dmijatovic:
I think we could ask @ctwhome  to come up with some ui suggestions on presenting this? It is important also to get clear understanding of the use case (what is the primary, secondary and nth goal of this page?). 

The number of KPI/columns currently is 14+. On the smaller screens this table might be challenging to present. For now we could add "full screen" mode icon and popup the table in the full screen mode. This will be useful for large screens.

ewan-escience:
Looks much better than my version, and with the new overlay function, there is also the option to see it without horizontal scrolling.

dmijatovic (approved):

jiskattema:
1. For any of the numeric scores (1,2,3, ..), it is hard to say if it's good or bad. however, a '0' is always bad. could a zero also be a red 'X' to make it stand out more?
2. Can the columns be made sortable? I have more than 50 projects, to have newest one on top (or ones with specific isssues) would be useful.;

PR #799, opened 2023-03-27
feat: suggest code platform based on existing entries

### LINKED ISSUE #781: Autosuggest platform based on repository url
When adding Software self hosted GitLab instances are not recognised as GitLab Platform, and the user has to set the platform themselves. In the context of the Helmholtz, many different GitLab platforms at the centres will be added to the RSD. A feature that auto-detects the platform based on the URL would be very helpful.

This could either be done in the admin panel, by adding the URLs and specifying their default platform, or by automatically generating this based on the user input. If a user adds a platform for the first time and set's the corresponding type, we could use this information for all future references.

cmeessen:
Fixes #781 

* automatically suggest a code platform based on currently publicly visible software entries
* if there exist several repositories referring to the same host, but set to different code platforms, the code platform with the most references will be suggested

How to test:

* `docker compose down --volumes && docker compose build --parallel && docker compose up --scale scrapers=0`
* login as normal user
* create a new software entry, e.g. using url `https://something.com/test1` and specifiy GitLab; publish the software entry
* create a new software entry and type (not copy paste) `https://something.com/test2` into the url field. The platform should be suggested as soon as `https://something.com/` is typed into the field; publish the entry

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/edit/information/AutosaveRepositoryUrl.tsx
+++ b/frontend/components/software/edit/information/AutosaveRepositoryUrl.tsx
@@ -1,7 +1,9 @@
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
+// SPDX-FileCopyrightText: 2023 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (dv4all) (dv4all)
 // SPDX-FileCopyrightText: 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2023 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 // SPDX-FileCopyrightText: 2023 Netherlands eScience Center
 //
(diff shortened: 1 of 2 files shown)


ewan-escience (approved):
Seems to work well. Maybe the database function could benefit from `SECURITY DEFINER` so that also unpublished software gets taken into account.

dmijatovic (approved):
Nice work!

Maybe is good to only take into a count entries from published software?

frontend/components/software/edit/information/AutosaveRepositoryUrl.tsx
```diff
@@ -1,7 +1,9 @@
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
+// SPDX-FileCopyrightText: 2023 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (dv4all) (dv4all)
```
fembau, line 4: @dmijatovic this line is duplicated ;) :thinking:

dmijatovic: Yes, it can cleaned now or later :-). Do what you preffer.

PR #805, opened 2023-03-27
fix: add software, project use null instead of empty string

### LINKED ISSUE #798: Subtitle of new project saved as empty string instead of null
When creating a new project, if you leave the subtitle field empty, the subtitle is saved as the empty string instead of `null`. When you clear en existing subtitle from the admin page, it is properly saved as `null` though.

dmijatovic:
# Use null on create software and project

Closes #798

* Use null for missing value when creating software
* Use null for missing value when creating project
* Subtitle is not required prop for software
* Min value is set to 3 charts
* Tests are updated

How to test:
* `make start` to rebuild app
* login as user 
* create new software without subtitle and verify that `short_statement` property has null value instead of ""
* create new project without subtitle and verify that `project_subtitle` property has null value instead of ""

PR Checklist:

*   [ ] Update documentation
*   [x] Tests

### DIFF
--- a/frontend/__tests__/SoftwareEditPage.test.tsx
+++ b/frontend/__tests__/SoftwareEditPage.test.tsx
@@ -97,7 +97,7 @@ describe('pages/software/[slug]/edit/[page].tsx', () => {
     expect(p403).toBeInTheDocument()
   })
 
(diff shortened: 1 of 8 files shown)


jmaassen (approved):
works as expected

PR #849, opened 2023-04-18
Use only theme colors in tailwind and mui

### LINKED ISSUE #823: Ensure use of theme colors in tailwind classes
The default tailwind config file offers more color options than we have defined in our theme. In out html markup there are few places where these non-theme colors are used (home page is one example). 

Using non-theme colors might create undesired theme issue for users of zelf-hosted RSD instances which want to apply their custom colors. 

The color options are shown by tailwind VS code plugin. We  can limit color options by moving `colors` object from `extend` section directly into `theme` section. The approach should reduce use of non-theme colors in tailwind css by their class names. 

In addition to tailwind configuration change we need to ensure (find) all non-theme colors used are added to our theme colors or replaced with an alternative color from our theme. The theme colors are defined in frontend/public/data/settings.json.

dmijatovic:
# Use only theme colors

Closes #823

*  We changed tailwind.config to use only theme colors. The tailwind plugin will suggest all possible color options.    

How to test:
* `make start` to build app
* navigate all pages for unexpected/different color use compared to live/dev version

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ctwhome:
That's correct, lw the background is indeed gray!

dmijatovic:
@ctwhome I have updated tailwind.config to extend the default color palette instead of using only theme colors (based on our latest discussion). I have added all tailwind colors used in new v2 pages to theme colors (settings.json). Can you please review?

ctwhome (approved):
It looks like it was expected. I like the comment on being careful and wise about adding custom colors to the theme settings.

PR #856, opened 2023-04-25
838 grid list selection highlights page

### LINKED ISSUE #7: POC fetch list of Dutch organizations
Find out what are the possibilities to list Dutch organization

*   [x]  Pull data from Surfsara (example: [list](https://www.surf.nl/over-surf/leden-van-surf))
  - I was not able to find an api endpoint which will return list of organizations, or search api
  - We might need to write a web scraper. Can we use ROR as alternative source?
*   [x] Pull data from ROR
    - For showing an list of possible research organizations from ROR database we cam use the endpoint: https://api.ror.org/organizations?query={{name}}
    - No API key is required, as shown in the [example on this page](https://ror.readme.io/docs/rest-api#query-parameter)
    - The information returned from the endpoint is [json with this properties](https://ror.readme.io/docs/ror-data-structure#ror-metadata)
    - I have not found any info about limited number of requests per second/minute in the documentation

## Conclusion

- ROR can be used to pull the list of all registered research organization. 
- The list seem to be extensive and it is expected to cover all research organizations "eligible" for adding to RDS
- ROR does not provide organization logo- 
- First impression is that we can use ROR id as unique id for research organization in our system
- Basic information received from ROR api (ROR-id and the organization name), needs to be enriched with our internal information if the company is already present in RDS.

### LINKED ISSUE #655: Behavior of filter for projects
For projects, I can now filter on both keywords and research domains (yay!). However, when selecting one of the filters, the other does not update the list of projects.

For example, selecting "Big Data (12)" as a keyword filters the project cards to the 12 that match, but the "Research Domain" filter still shows the full list. This is not really intuitive, as there is a high chance of me picking a research domain which has 0 big data projects. 

It would be better if the lists are linked somehow? Updating one reduces the list of the other to those that match?

### LINKED ISSUE #789: Match filters counts with selected options
related to #655 
The current filter panel shows the number of keywords, program languages, and licenses that exist for all the software. 
Is nice to have, but also gives a better indication of what types of results are available base on a filter selection.

It is probably needed to create an RPC that returns the filtered counts the based on:
- search input
- Keywords (including the keywords list counts itself)
- Programm languages (idem)
- Licenses (item)

Example: 
- After selecting the keyword "GPU", the rest of the counters should be updated with taking in account the new filter
image

### LINKED ISSUE #838: Add the possibility to display the software cards by Grid or List mode
Show the software list on a list: 

Design
- [ ] Visual Design

Implementation
- [ ] List implementation
- [ ] Store the decision on localStorage

for example: 
image

image

dmijatovic:
# Enable grid or list selection on highlights page

Closes #655 
Closes #789
Closes #838

* Requests on highlights page are done on server side, including dynamic filter lists
* The filters reflect current software selection. The options that are not possible are not shown in the filter options
* Position message shows `Page 1 of XXX results`
* Choice between 3 layouts (see images 2,3 and 4): masonry, grid and list. The choice is saved in the cookie `rsd_page_layout`
* Choice between 12, 24 and 48 items on the page. The choice is saved in the cookie `rsd_page_rows`.
* Keep page scroll position when selecting/changing filter.

How to test:
* `make start` to rebuild app
* navigate to [highlights page](http://localhost/highlights) (not in menu)
* test different filtering and the counts shown in the filter

## Example highlights page
image

## Example masonry card layout
image

## Example grid layout
image

## Example list layout
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ctwhome:
I love the ending result, wow, I think it adds so much value to the search. 
I am not sure when the user would click on the number to show a different number of items per page, but I guess it doesn't harm anyone. It is also not intrusive.

A few small tweaks: 

List avatars. Maybe not show them if there is no picture. Also instead of a rounded image, a square would make more sense since the cover images are also squared
image

The cards in the grid mode with no cover image create a bigger space at the bottom
image


The list items should be all clickable not only the title, but also not change the color of the list item to green when hovering, to keep consistency with the other views. The row should be clickable, not just the title:
image

ctwhome:
A possible enhancement:

when clicking on the highlights page (future software page) the view type and the number of cards to display are not saved in the locastorage of the user, meaning that they get reset when they go out of the page. 
The url saves the page nd the rows to show, which is useful for sharing the link, but they get lost when exiting the page
urL `?page=1&rows=24`

We could save the view type (masonry, grid, list) and the number of cards in local storage so the use only have to choose the option once (at least on that computer)

dmijatovic:
@ctwhome Thanks for your feedback. The improvement suggestions are implemented. Can you double check and approve?

ewan-escience (approved):
Nice work, I especially like the layout options.

In the future, I think it would look better that when software does not have an image, that in the grid and list view, we put whitespace where normally the image is.

PR #870, opened 2023-05-09
fix: reset publish switch value on failed update

### LINKED ISSUE #867: software publish switch should respect error status from API call
I'm working on setting up constraints a software entry must have before it can be published. I created a database check function which is triggered on modifications of the software table. The function raises an exception then.

#### Current behavior:
Switching the "published" switch on `software/xyz/edit/information` will show an red error message box and the switch changes to activated state.
Reloading the page will correctly show the deactivated state.

#### Expected behavior:
The switch should go back or keep deactivated state.

dmijatovic:
# Reset publish switch value on failed update

Closes #867

* reset publish switch value on failed update for software and project item
* renamed software component to AutosaveSoftwareSwitch    

How to test:
It is not possible to test without temporary changing something in the code to cause an error :-(. Maybe @fembau can test it on his branch developing new feature that requires this adjustment?

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/projects/edit/information/AutosaveProjectSwitch.tsx
+++ b/frontend/components/projects/edit/information/AutosaveProjectSwitch.tsx
@@ -43,12 +43,14 @@ export default function AutosaveProjectSwitch({project_id,name,label,defaultValu
         defaultValue:value
       })
     } else {
-      showErrorMessage(`Failed to save ${name}. ${resp?.message}`)
+      showErrorMessage(`Failed to save ${label.toLowerCase()}. ${resp?.message}`)
       // On error we reset value to defaultValue/original value
       // to let user know that update failed
-      resetField(name, {
-        defaultValue
-      })
+      setTimeout(() => {
+        resetField(name, {
+          defaultValue
+        })
+      },100)
     }
   }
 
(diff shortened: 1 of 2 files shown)


ewan-escience:
You can test it by applying the following SQL to the database

```SQL
-- create a trigger that rejects changes to is_published
CREATE OR REPLACE FUNCTION z_reject_published() RETURNS TRIGGER LANGUAGE plpgsql AS
$$
BEGIN
	IF NEW.is_published IS DISTINCT FROM OLD.is_published THEN RAISE EXCEPTION 'You are not allowed to edit the publication status'; END IF;
	return NEW;
END
$$;

-- apply the trigger on updating software
CREATE TRIGGER z_reject_published BEFORE UPDATE ON software FOR EACH ROW EXECUTE PROCEDURE z_reject_published();

-- drop the trigger to restore normal behaviour
DROP TRIGGER z_reject_published ON software;

```

ewan-escience (commented):
It works, but I think it would be better to download the new is_published status after trying to change it instead of blindly changing/resetting the status.

fembau:
I works in my other feature branch.

In the failing case the publish switch now has a somehow none to flashing animation and is maybe confusing for the user. I suggest to let the switch-on animation finish before starting the switch-off animation..

fembau (approved):

dmijatovic:
@fembau I added small delay when reseting the switch to original value and improved error message a bit. In the error I use property label instead of name. Can you let me know if this version is sufficient for your needs?

PR #873, opened 2023-05-12
Software overview v2

dmijatovic:
# Software overview v2

* Software overview with highlights, dynamic filters and view options for masonry, grid and list
* Page accessibility improvements based on Lighthouse audit suggestions
* Separate highlights page is removed
* Old software overview page is removed
* Unit tests created for new software overview page    
* Linting warning for use of img markup is turned off for the whole project
* BREAKING CHANGE comment is added to commit to bump major version number

How to test:
* `make start` to rebuild app
* login as rsd admin (professor3 if you have same .env values as me)
* navigate to admin - highlights section. Examine defined highlights. Add more or remove items.
* navigate to software (overview) page.
* confirm highlights are shown. When you remove all highlight items in the admin section, the highlight section will not be shown at all
* test order and filters on the left side of the page. After selection of an filter item the remaining filter options will be updated to remaining options present in the remaining selection. 
* test view options: masonry, grid or list. Change number of items per page.
* test pagination at the bottom of the page to navigate between the pages. Note! When using pagination the page will scroll to top.

## Example masonry
image

## Example grid
image

## Example list
image

PR Checklist:

*   [ ] Update documentation
*   [x] Tests


ctwhome:
Let's goo!

ctwhome (approved):

jmaassen:
Some initial comments from my side. We can convert these to issues and merge the PR anyway.

- The images do not work well in the grid layout. With the generated examples, almost all of them are cropped in a strange way. I think we should make the images vertically bigger in grid layout so they becomes square, just like in the highlight bar and the list view. 

- The cards seem unbalanced at the bottom, between the keywords and the icons there is a lot of vertical whitespace. Should there be some extra info there? Maybe the harvesters haven't run? Otherwise, move the keywords down?

- Get rid of the 12/24/48 selector and just pick a default that makes sense for the layout selected

- I have a scrollbar under the highlights, which doesn't look nice. Would dots be an option? Or remove it all together

- The highlights look weird when on the first or last item (due to the empty space next to it). Can we do infinite scrolling instead, like Huygens does: https://www.huygens.knaw.nl/resources/

- With a large number of keywords, the "+x" has an offset which looks a bit strange

- When adding highlights, it's not really clear at what position the new highlight is inserted? It seems to insert in the middle sometimes?

(to be continued)

cmeessen:
I think it would be nice to have more options for ordering:

* newest
* alphabetically

dmijatovic:
@cmeessen I added order by Name in this PR. The newest needs more thinking about it. It could be "latest added to RSD", "recently updated in RSD", recently commited to software repo (if we have this repo url) and recently created repository (or time or first commit). Most of these are useful in some situations :-).

cmeessen:
>  The newest needs more thinking. It could be "latest added to RSD", "recently updated in RSD", recently commited to software repo (if we have this repo url) and recently created repository (or time of first commit).

Good point. I was referring to "newest additions" in terms of switching the "published" switch. But then again, one could jump to the front of the list by switching it on and off.

PR #874, opened 2023-05-16
Change content sample images

ctwhome:

* Changed the content of the sample software images to give it a closer look at scientific software packages
* The images preserve the screen ratios and extensions (png, jpg and svg) like the original ones to keep the diversity.

image


jmaassen (approved):
Seems to work as expected

PR #892, opened 2023-06-01
fix: add a request timeout on all HTTP requests

ewan-escience:
# Scrapers request timeouts

* All HTTP requests of the scrapers now have a timeout, which is triggered when the headers of the request have not been received by that time.

How to test:

* Check that the scrapers still work
* Check in the code that I didn't forget to add a timeout somewhere

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/scrapers/src/main/java/nl/esciencecenter/rsd/scraper/Utils.java
+++ b/scrapers/src/main/java/nl/esciencecenter/rsd/scraper/Utils.java
@@ -20,6 +20,7 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.Base64;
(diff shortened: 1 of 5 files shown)


dmijatovic (approved):

ewan-escience:
For context, this PR was made as a response to an incident at Helmholtz, where the scrapers showed a memory leak. In their logs, many requests generated exceptions, (which is OK, since that will eventually lead to the scraper being done and exiting). We couldn't find the actual casue of the memory leak, but we suspect that some HTTP requests kept hanging forever, causing some scrapers to never exit, while cron would just keep on spawning more processes.

We now added timeouts that are fired when the headers for an HTTP requests have not been received yet. They don't trigger, however, when the request body is hanging. We hope that the current approach is sufficient, but if it is not the case, we should take extra measures to ensure that each scraper process is closed after a certain time limit.

cmeessen (approved):

PR #903, opened 2023-06-07
Replace postman tests

### LINKED ISSUE #813: Move away from Postman for API tests
We should replace the Postman/Newman tests ([here](https://github.com/research-software-directory/RSD-as-a-service/tree/main/backend-postgrest/tests)) with something else, for two reasons:

1. Since February 2023, you can only do 25(!) collection runs per month on a free account ([announcement](https://blog.postman.com/postman-2023-product-and-versioning-updatesand-vs-code-extension/)).
2. The Postman interface is becoming less suitable with the increasing number of tests.

To address the second point, we should go for a full code-based solution. Considering the languages already in use, we should either go for Java or JavaScript/TypeScript.

ewan-escience:
# Replace Postman tests

* Postman tests are replaced by full code tests written in Java as a Maven project
* Main libraries are [REST Assured](https://rest-assured.io/) and [JUnit 5](https://junit.org/junit5/)
* Not all tests were migrated
* Add deprecation note on the old README but keep the folder so the old tests can still be found

How to test:

* Read the new `README.md` and check if the instructions are clear enough for you to be able to run the new tests 😉
* Check if any essential tests are missing

Closes #813

PR Checklist:

*   [x] Update documentation
*   [x] Tests

### DIFF
--- a/.github/workflows/all_tests.yml
+++ b/.github/workflows/all_tests.yml
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all)
-# SPDX-FileCopyrightText: 2022 - 2023 dv4all
 # SPDX-FileCopyrightText: 2022 - 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 # SPDX-FileCopyrightText: 2022 - 2023 Netherlands eScience Center
+# SPDX-FileCopyrightText: 2022 - 2023 dv4all
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -35,11 +35,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - name: "newman tests with docker compose"
-        working-directory: backend-postgrest/tests
+      - name: "backend tests with docker compose"
+        working-directory: backend-tests
         run: |
           docker compose up \
-            --abort-on-container-exit \
             --exit-code-from postgrest-tests
 
   scrapers-tests:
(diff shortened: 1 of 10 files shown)


jmaassen:
Looks good and seems to work well. Some questions/comments:

- On the console I see a few errors (`ERROR:  permission denied for table software`), but I assume this is exactly what the test is trying to achieve?
- In the past I used Jacoco to generate coverage rapports after testing, would also be nice to use here? It can produce a set of local HTML files, but I also use it in combination with services like Codecov and SonarCloud. See for example:
   - https://www.jacoco.org/jacoco
   - https://app.codecov.io/gh/xenon-middleware/xenon/tree/master/src%2Fmain%2Fjava%2Fnl%2Fesciencecenter%2Fxenon
   - https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=xenon-middleware_xenon
- I often use Palentir to directly control docker-compose from the unit test itself. Not sure it this is needed here? See:
   - https://github.com/palantir/docker-compose-rule
- Can we also use RestAssured to mock external services? Like datacite, github, etc? By replaying replies (especially weird corner cases) we could improve our test coverage.

jmaassen (approved):

ewan-escience:
Thanks for the feedback @jmaassen!
- Indeed, some tests *should* produce errors on the backend. As long as all the tests succeed it is probably ok.
- Since we're testing an "external" API (the PostgREST  backend and the database), code coverage is not applicable here, unless we can somehow configure JaCoCo to measure the code coverage of the database initialisation files.
- I'll have a look at Palentir as that could be interesting. The tests now assume that the necesary external services are already running and it polls every second (for 30 seconds max) if the backend is available.
- REST Assured is an HTTP client only. For mocking responses we could look at e.g. [Mockito](https://site.mockito.org/) to directly mock the responses from REST Assured or at e.g. [MockServer](https://www.mock-server.com/) for setting up a full mock server.

Issue #923, opened 2023-06-26
Add support for categories for software and projects.


fembau:
Add support for categories for software and projects.

Related to #741 but starting with a more generic approach and simplified requirements.

Issue #915, opened 2023-06-19
improve code quality by CI linting pipelines


fembau:
Add a CI pipeline which calls ESLint on each commit.

Additionally use p.ex. https://prettier.io for the frontend and [focus](https://prettier.io/docs/en/why-prettier.html) on coding instead of formating. Prettier could also be integrated with ESLint.

jmaassen:
Agreed. We should also do this for other languages (like the Java in the backend).

fembau:
and SQL

ewan-escience:
I'm trying out Prettier for Java and SQL files. You can see the differences here: https://github.com/research-software-directory/RSD-as-a-service/compare/915-code-quality?expand=1.

It's not all better, look for example at lines 43 -- 45 of authentication/src/main/java/nl/esciencecenter/rsd/authentication/OrcidLogin.java. It's apparently not good at handling if statements without curly brackets.

ewan-escience:
Here is a screenshot of the code I meant in my previous comment, as the code will change:

![Screenshot_2023-08-07_11-17-00](https://github.com/research-software-directory/RSD-as-a-service/assets/91328544/61e53d2e-f6e1-4f1e-9923-25b456a1ea9f)

I will be trying out other formatters/linters, e.g. [Checkstyle](https://checkstyle.org/) for Java and [SQLFluff](https://sqlfluff.com/) for the database.

dmijatovic:
For the frontend I find current eslint sufficient. It runs during the build and can be runned locally from the frontend folder. I do lint check regularly manually. We can add frontend linting check to CI though (`yarn lint`).

Adding Prettier to the chain requires additional dependency and VS plugin to be installed by fe developers. I would prefer lighter approach with eslint only.

fembau:
Prettier could be [integrated with ESLint](https://prettier.io/docs/en/integrating-with-linters.html). I used this setup before. No extras for the IDE necessary, ESLint is enough.

Using prettier is very comfortable. Currently I have to switch it off because when making changes of course to whole file gets reformated and this makes it hard to understand PRs.

fembau:
I found an issue in my VSCode setup and now ESLint is working as expected :)
Also I understood now that ESLint is used as formater.

PR #946, opened 2023-07-20
923 add support for categories for software

fembau:
# add support for categories for software

Addresses #923 :)

This is a first functional prototype. This branch also includes some [pre-defined categories](https://github.com/research-software-directory/RSD-as-a-service/blob/923-add-support-for-categories-for-software-and-projects/database/999-add-hgf-categories.sql) for testing.

### Some intro

- categories are organized in a hierarchical structure (tree)
- naming:
  - a **category** is a leaf in the category tree
  - a **parent category** is a tree node with sub-categories
  - a **top-level category** is a tree node with no parent
  - a list of categories traversing the category tree from the top-level category down to a leaf is the **category path** of the leaf
  - the **visual representation of a category path** could be p.ex.
    `top-level cat / sub-cat / category`
- software/projects could be related to multiply categories
  - side effect: because categories are the tree leafs, categories could be re-structured without harming the association to software/projects
  - POV topics are related to research fields (see list of pre-defined categories), thus the origin structure would be:
    `POV phase / research field / topic`
    ➜ design decision: for now the middle level is left out:
    `POF phase / topic`

### Version 1

- Minimal functionality (!)
- maintainer:
  - .. is able to manage categories for a software:
    - categories = research fields and POF topics
    - start with really simple UI: add/remove button, list of (all) categories
- admin:
  - no admin interface for categories yet. Data for pre-defined categories is injected by an extra *migration file* during deployment
- user view of a *software*:
  - add a new headline "Categories" (+icon) in the right lane above "Keywords"
  - ...and a (sorted) list of category paths, p.ex. like this (see example):
    - Research fields | Energy
    - POF IV | Energy System Design
  - show *long name* on mouse-over an item in a category path

#### Testing
- [ ] add backend test?! how?
- [x] add frontend test?!

#### Usability / Features
- [x] find good representation of category paths
  - p.ex. how should we abbreviate some of these long category names?
  - design
  - interaction (p.ex. show long name on hover)
- [x] add category: do not show already added categories in the offered list

#### Code structure
- [ ] the API calling functions have a lot of copy&paste code. Let's create some generic helper functions for API calls

---

#### How to test:
*

#### PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


frontend/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
fembau, line 3: enables Array.entries()

ewan-escience (commented):
Here are a few things I noticed when looking at the database files:

- there is no row level security for `category`, which means that any logged in user can create, edit and delete categories
- there is no check that the categories cannot form a cycle (e.g. cat1 has parent cat2, cat2 has parent cat1)
- please add the corresponding `sanitise` functions to `category`
- `is_published = TRUE` can be shortened to `is_published` in a `where` clause
- some keywords are not capitalised, e.g. `references`, `as`

We can furthermore add tests to the `backend_tests` module to test if all functions and checks work as intended, let me know if you need help with that.

jmaassen:
Some comments from the UI perspective: 

- The design of the categories is now quite different from the keywords directly below:

Screenshot 2023-07-25 at 08-11-04 Edit software Research Software Directory

I think it would be better to use the same design approach: rounded boxed and an "(x)" for deletion instead of a delete button?

- When adding a category, the top entry is "click to add more categories" which implies clicking it will allow you to define a new category (which isn't the case).

- The list of categories becomes quite unwieldy when you have many categories of deeply nested ones. Would an approach similar to Research Domains in projects work?

Screenshot 2023-07-25 at 08-12-59 Edit project Research Software Directory

- When adding multiple categories with the same root, the full path is always shown in each box:

Screenshot 2023-07-25 at 08-24-31 Software Country innovative Chromium

for research domains in projects we use a slightly different approach: 

Screenshot 2023-07-25 at 08-15-25 Project Administrator District Koruna

In this case, LS is the parent of both LS3 and LS6 and grand parent of LS3_14. The full path isn't shown for the last three. Would a similar approach work for categories? This may depend on the rules there are to select one? On difference I see is that for Research Domains, one is also allowed to select a top level or mid level one, not just a leaf one.

ctwhome:
A small comment on the functionality. When the user is not able to add categories (I believe that is because of privileges) there is still a message saying "add categories", but it is not possible to manually add any. It gives the illusion that new categories can be created, but no.
image
image

Also, maybe is wise to use muy components instead of native HTML select to keep design consistency and functionality as Jason mentioned.

ctwhome:
In terms of design I think is oke, the whole information page has become too full of features that in my opinion needs an aesthetic redesign to keep the form inputs more organized.

fembau:
About @ewan-escience comments:
- row level security for `category`. In the first place added a read-only rule for everyone. A admin/maintainer interface is introduced in another issue #952 
- added constraints to prevent cycles.
- added sanitizer to prevent set category `id` (like other tables). This also prevents bulk insert of cycles.
- code formating, see #915
- tests -> TODO

fembau:
About @jmaassen and @ctwhome comments:

Compared to the categories used for projects, my idea of categories is that only a _category_ (see definition of terms above "naming") aka a tree leaf is associated to a software/project. So having relations between the software and all the sub-level category entries makes no sense for me.

The functional prototype enables us to play around with hierarchical nested categories and tackle the arising issues.

One UI challenge is how to represent category (path) names especially the longish or deep nested ones. The available UI components are not satisfying, p.ex. using an ellipse will hide the end of a category path which is the most important information for the user.

Selecting a category is another challenge. Currently a simple dropdown is used. As proposed something like the project categories (level 1,2,3) thing could be used, but starting with one dropdown and dynamically appearing new dropdowns for each sub-level. Reaching a tree leaf could finally display a confirm button.
Or a [cascading menu](https://sintel.zendesk.com/hc/user_images/2LC8952ynP5uG-n6SHvRLg.png).

I added some more examples for category path representation using available UI components and also some more example categories to play around with deep nested and longish category names.

ewan-escience:
The frontend Docker image doesn't build for me.

dmijatovic:
@fembau few suggestions for validation. You can use following commands in the `frontend` folder to run checks locally. It does requires npm, yarn and installation of all dependencies (use yarn install)

- `yarn lint` to check for linting errors. In the current code, beside warnings there are 2 errors on missing key props in SoftwareCategories.tsx
- `yarn test` to run test. 
- `yarn build` to build frontend. Note the frontend will not build if there are linting errors (warnings are ok).

Maybe we can have a look at the code editor you use? If you use VSCode and have required plugins installed the linting errors should also be shown in the code editor. I expect similar kind of plugins are available for other editors too.

fembau:
**About representation of categories:**

As discussed the first implementation should have:
- Instead of having a headline "Categories", each top-level category should be represented by it's own headline.
- only consider a hierarchy depth of 2

ctwhome:
After a design session between @fembau and me, we came up with a proposal to display the categories section in the following way. This design allows to use of only one component either to display in the view mode or admin mode to improve user experience and combine development efforts. 
Note that this approach assumes that there is an existing list of categories defined by the organization, of which there is not a design yet nor it is outside the scope of this PR.

image

fembau:
As discussed with @ewan-escience we put the metadata of categories, like color, icon, etc. into the same database table. Most metadata columns would be empty. Other options we talked about: put metadata into extra table (to save space); add a column of type json to store metadata.

fembau:
closing in favor of #984

PR #967, opened 2023-08-14
chore: exclude test and mock files from sonar code analyses

dmijatovic:
# Exclude folders and files from Sonar analyses

* Do not analyse test, mocks and assets folders
* Define all test files   

How to test:
* De sonar analyses should now exclude large number of irrelevant files from it's analyses.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/sonar-project.properties
+++ b/frontend/sonar-project.properties
@@ -1,3 +1,4 @@
+# SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
 # SPDX-FileCopyrightText: 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 # SPDX-FileCopyrightText: 2023 Netherlands eScience Center
 #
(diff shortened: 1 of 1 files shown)


ewan-escience (changes requested):
You need to use `sonar.coverage.exclusions` instead of `sonar.exclusions`, see the section `Refined scope settings` on https://docs.sonarcloud.io/advanced-setup/analysis-scope/.

dmijatovic:
> You need to use `sonar.coverage.exclusions` instead of `sonar.exclusions`, see the section `Refined scope settings` on https://docs.sonarcloud.io/advanced-setup/analysis-scope/.

Yes, I saw that. The exclusion I defined is not only for coverage but for all sonar analyses. These folders should folders should be excluded from Sonar analyses completely not only for the coverage. 

For the coverage to be identical to local Jest  report we would need to use inclusion option but that option does not exists in Sonar for the coverage. I am accepting that there will be some difference :-(. I expect if the test files are defined properly the stats will improve. From there we can decide on next steps.

ewan-escience (changes requested):
On the same page, it says

> The test scope parameters (source.tests , sonar.test.exclusion, and sonar.test.inclusion) do not have anything to do with setting up test coverage reporting (see [Test Coverage](https://docs.sonarcloud.io/enriching/test-coverage/overview/)).

So you will not get the desired result this way. It will be worse, because the test code should be analysed but not be in the test coverage.

ewan-escience (approved):
Let's see if it works...

PR #970, opened 2023-08-28
Global notifications

cmeessen:

* administrators can create global notifications via the admin interface, e.g. to notify visitors about upcoming downtimes

How to test:

* login as rsd_admin to be able to access admin section
* navigate to Anouncement section of admin page
* Fill in info in the form and hit save button
* Reload RSD to view notification. It will be shown on every page. After using X button it will be hidden. This flag is only in memory and if you reload RSD the notification will appear again.
* The notification uses `warning` colors defined in the settings (light theme).

## Example admin section
image

## Example announcement
image

PR Checklist:

*   [ ] Update documentation
*   [x] Tests


dmijatovic (approved):
Look good as first (minimal) version of the feature.

jmaassen (commented):
Works well from a user (admin) perspective.

ewan-escience (commented):
A few comments/suggestions, but non are deal breaking, so let me know if and what you want to do with them.

1. Since the text is set by admins only, consider removing the 200 char limit.
2. In the frontend, I cannot fully remove the text, I can only make it invisible, is this intentional?
3. Consider enforcing allowing only one row to exist in this table, see e.g. `017-create-narcis-table.sql` for how to to this, this also allows you to drop the UUID column.

cmeessen:
@ewan-escience thank you for the suggestions. I applied the database trick which is quite neat. As for the text, I think it does make sense to not be able to completely remove it in the frontend. Otherwise, one might accidentally end up with an empty notification bar.

If there is nothing else, this PR is ready to be merged.

Issue #975, opened 2023-09-01
Dynamically load icons


fembau:
To use icons it is currently necessary to import them individually, p.ex.:

    import CategoryIcon from '@mui/icons-material/Category';

For categories #923 we want to have an icon in the category headline and the icon name is not known at build time. We need a mechanism to load icons on demand.

fembau:
extend `frontend/components/typography/Icon.tsx`

Issue #952, opened 2023-08-01
add interface for managing categories as admin


fembau:
Related to #741, #923

- Define user journeys and use-cases for categories and how to manage/organize them.
- Create a user interface for managing categories as admin

ewan-escience:
Closed by #1227.

Issue #741, opened 2023-02-21
Customisable pre-defined category lists


cmeessen:
We have the following use case at Helmholtz: there exist six different overarching [Research Fields](https://www.helmholtz.de/en/research/research-fields/). Each centre dedicates it's research to at least one of the research fields, and they are important when it comes to reporting, proposals etc. We would like to be able to indicate the Research Field a software or project belongs to in a prominent manner.

This could be fulfilled by adding customisable category lists to the RSD. The lists are maintainable in the frontend by Administrators of an RSD instance, and can be used by any user in the instance. Here are some drafts how it may look like in the RSD:

#### Admin frontend

![image](https://user-images.githubusercontent.com/14222414/220318249-d884daba-1ae2-4cac-ac94-f22787292596.png)

* "Visible per default" indicates whether the category is always shown on the edit page. This will nudge users towards actually using them.

#### Software-edit page sidebar

![image](https://user-images.githubusercontent.com/14222414/220318460-0de603ce-95b0-4a4e-8422-188f201a221b.png)

* Display the categories above the keywords to indicate that they may be more important as they are predefined lists
* If other categories exist, users can add them via a dropdown menu

#### Software page

![image](https://user-images.githubusercontent.com/14222414/220318961-423455b6-8247-4d15-83c7-7847a263c9bb.png)

On the software page, the keywords should be shown above the keywords as well.

Suggestions are welcome

ewan-escience:
Some implementation considerations:
* When making a category mandatory, most existing pages won't have a value yet. Therefore, it is impossible to enforce this in the database. However, we can encourage/enforce this in the frontend whenever a maintainer tries to edit a page.
* If an admin _unchecks_ the `multiple selections allowed` box, what should happen for existing software that already have multiple values? Should we delete all values for that software? Or all but one (but which one, this makes the least sense to me)? Or should we not enforce it in the database but in the frontend only?

cmeessen:
Related to #616

dmijatovic:
I also have some implementation considerations. This section is now on the first page in the right panel (see image). There are ideas to simplify first page by moving all items except Publish and License into a Software metadata page. The idea behind is to link ConceptDOI with the imports. Introducing wizzard is also an alternative. The logo will be moved into the Image page. Thinking about alternative location of this feature might be beneficial.

## Current page info layout

[...]

PR #984, opened 2023-09-11
923 add categories for software

dmijatovic:
PR #946 was polished and prepared for v2 -> created this new PR.

### How to test:
remove extra file extension from `999-add-categories.sql.example`, optional edit file and finally build application as usual.

On a software edit page you will find a new headline "Categories".


fembau:
@dmijatovic it seems that only you can switch off draft state ;) Please also review this PR (can not add you)

dmijatovic:
@fembau Shell we close the other PR in favour of thisone then?

ewan-escience (changes requested):
Nice work! I have a few small comments/questions:

- What is `frontend/.yarn/releases/yarn-1.22.19.cjs`? It is 5MB in size. Can it be removed? Same for `frontend/.yarnrc.yml`.
- Please remove the `AND is_published` from the policy 'anyone_can_read' on `category_for_software`, that logic is already covered by the policy on software.
- Please capitalise 'as' everywhere in `category_paths_by_software_expanded`
- In the test, please put the `User` class in a dedicated file.
- I'm not sure about the `User` constructor using HTTP requests to initialise itself. I think it's better to have a static factory method somewhere that does this for you and returns a `User` upon success. But this could be done later as well.
- Line 273 can be deleted as it only contains a unnecessary semicolon.
- Maybe a test can be added that tries to create a cycle in the categories.
- Is it intentional that there is not an admin interface to add/edit categories?

jmaassen:
Works well, some comments: 

- I was slightly confusing at first that the categories in the drop down list when selecting are different that what you see after selecting. This was mainly because in the example the "Category X aka X.Y.Z" is shorter than the short name ;-)

- having an icon is a nice touch. Do we have documentation on which icon names are valid? Currently it says "science".

- for consideration: can we use the same approach for projects? You could say that "Research domains" is simply the top level category.

dmijatovic:
@fembau If you need the yarn settings @ewan is referring, please add these to .gitignore file. That way these will not be included in the repo.

fembau:
Thank you for your feedback! :)

@ewan-escience @dmijatovic:
- ejected yarn topic from this PR. see also [discussions/843](https://github.com/research-software-directory/RSD-as-a-service/discussions/843)
- _Please remove the AND is_published from the policy 'anyone_can_read' on category_for_software, that logic is already covered by the policy on software._ --> This should prevent reading information related to not-published software.
- moved `User` to separate file + added factory. Also created a `class Commons`
- beautified code :)
- added test for cyclic categories.
- admin interface is postponed to #952

@jmaassen:
- improved text of example categories :)
- select leafs only: yes, as proposed in the related issue I think this is the proper way of using a catalog/vocabulary here... Anyhow it is currently not prevented by the backend...
- icons: until we have #975 available icons are defined in [frontend/components/typography/Icon.tsx](https://github.com/research-software-directory/RSD-as-a-service/pull/984/files#diff-8fee6e327b2bad51af55a1fa895c339f5c848709253edf586cc0ba0286878013). I put a note into the examples file.
- also for projects: yes, this is intended. Kept this out of this PR to make it smaller and setup a generic foundation first ;)

We have already seen that generic categories inspire lots of ideas what we could do next :)

ewan-escience:
> This should prevent reading information related to not-published software.

But that is already covered by `SELECT 1 FROM software WHERE id = software_id`, which takes the published status into account because of the row level security on software.

A small pet peeve of mine, the example SQL file is indented with spaces instead of tabs. :)

That's all from me, nice work!

cmeessen (approved):
Awesome, this is looking really good already.

I created #992 to address one aspect that was part of #741 but was not implemented yet.

ctwhome (approved):

ewan-escience (approved):

PR #996, opened 2023-09-28
Hide masonry layout

dmijatovic:
# Hide masonry layout in software overview

* The images currently used in the software pages are less suitable for masonry layout. Therefore we decided to remove masonry layout option in the first v2 release. 
* In addition, some refactoring is performed to share layout options between software and project overview pages because the options are currently identical. 
* The layout jump on software/project overview page when using list view (can be seen on dev) is reported by Jesus. This "layout jump" is addressed in this PR too.
* Tests are updated

How to test:
* navigate to software overview page and confirm that only two options are present: grid and list
* navigate to project overview page and confirm that only two options are present: grid and list
* navigate to organisation page and confirm that same options are present on organisation software and project pages
* select list view and navigate between software and project overview pages. During the changes the list view layout should be maintained without jumps or switches to some other layout option.

## Software overview
image

## Project overview 
image

## Organisation overview
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ctwhome:
@jmaassen @maaikedj @ewan-escience 

I suggest removing images for now for software cards to avoid displaying many of them with a gradient placeholder until the image/logo/cover software is figured it out. 

The Design intention for software images was never intended to display logos, but a more curated cover image. The current solution is polluting the image cards with either a placeholder or low-resolution logo which disrupts the new software card design. 

Software cards without logos would look like the following. In my opinion much cleaner and less disruptive. 
image

maaikedj:
> @jmaassen @maaikedj @ewan-escience
> 
> I suggest removing images for now for software cards to avoid displaying many of them with a gradient placeholder until the image/logo/cover software is figured it out.
> 
> The Design intention for software images was never meant to display logos, but a more curated cover image. The current solution is polluting the image cards with either a placeholder or low-resolution logo which disrupts the new software card design.
> 
> Software cards without logos would look like the following. In my opinion much cleaner and less disruptive. 

Agree

jmaassen:
After further discussion with Christian and Maaike we decide to release with this view after all. It works well on the Helmholtz instance, so we believe that it should also work for eScience if we improve the data quality.

dmijatovic:
We decided to take different approach

PR #1004, opened 2023-10-02
Contact person improvements

### LINKED ISSUE #885: Contac persons appears twice in the team section
Peter Kok is repeated. 
https://research-software-directory.org/projects/emotion-recognition-in-dementia

image

### LINKED ISSUE #948: It is possible to make contact person to more than one member in projects
To change the contact person of a project, you have to first change the current contact person. 
I would expect that the new contact person switch will disable the previous one by default.
When selecting a new contact person there is no indication or information that now there are 2 contact persons, and it is unsure which one is selected.

To fix: 
- when selecting a contact person, disable the rest in that operation, leaving the new contact person as the only enabled option.

ewan-escience:
# Contact person improvements

* Contact persons are not shown in the regular list of contributors anymore.
* Each contact person is now shown as such, instead of only one (which was the first one returned from the API).

How to test :

* Login as admin
* Edit a software page so that multiple people, up to everyone, is a contact person, inspect the page
* Make it such that no one is a contact person, inspect the page

I suspect that this might spark some discsussion, go ahead. 😉

Closes #885

Closes #948

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/ContributorsSection.tsx
+++ b/frontend/components/software/ContributorsSection.tsx
@@ -1,25 +1,24 @@
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 dv4all
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 // SPDX-FileCopyrightText: 2023 Netherlands eScience Center
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import {Person} from '../../types/Contributor'
+import {Contributor, Person} from '../../types/Contributor'
 import PageContainer from '../layout/PageContainer'
 import ContributorsList from './ContributorsList'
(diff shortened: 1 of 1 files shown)


dmijatovic:
@ewan-escience I had a look at it. Technically it works but we should change design if we want to support multiple contacts. 

image

I would advice to design new contributor card, have a contact person badge (or some other visual "sign") and show these at the top of the list. This will make layout lot easier to manage. The current design is not optimal for having multiple contact persons in my opinion. I am interested to hear what others think @jmaassen, @ctwhome, @cmeessen .

ewan-escience:
I agree, but I think that in production most pages have 0 or 1 contact persons.

dmijatovic:
I propose to decide first on the following issues and create design to optimally cover the needs.

- **decide whether we need to have contact info or contact person(s)**. I see a difference between contact info and contact person. The contact info should be a single instance in my opinion and should have own section on the software page, separate from the contributors. 
- **if we choose to have contact person(s)** then I suggest two separate sections (one for contributors and one for contacts). One person can be contributor and the contact person, I see these as two separate "roles".

@jmaassen @ewan-escience @cmeessen @ctwhome @fembau What do you think?

PR #925, opened 2023-06-28
Add closed_source flag to database

dalonsoa:
# Pull request Title

As part of the ongoing discussions to contribute to the RSD, here is the first step towards adding explicit support for open/close source software.

Fixes https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/12

Changes proposed in this pull request:

* In this case, it is a simple flag in the DB to indicate if a piece of software is open source or not. This flag could then be used in the front end (down the line) to require mandatory fields for open/close source or to show/hide certain parts of the page.

How to test:

* (Not testable, yet)

PR Checklist:

*   [ ] Increase version numbers in `docker-compose.yml`
*   [ ] Link to a GitHub issue
*   [ ] Update documentation
*   [ ] Tests

## Question:

I understand that there should be a migration in the dB in order to apply these changes. However, it is not clear to me if the instructions you indicate in the [relevant README](https://github.com/research-software-directory/RSD-as-a-service/tree/main/data-migration) should be used or no. Could you advise on how to proceed? I'll proceed based on what you tell me and mar this PR as a ready to review.

PR #1012, opened 2023-10-11
Cs responsive form

cc-a:
Following on from #925 (which added a database flag for closed source software packages) we've been exploring the UI changes that would work with this to actually allow users to submit a closed source package. This PR is presented in draft form to provide a starting point for discussions on the UI changes. The implementation and the appearance of the UI elements is preliminary.

- Adds a switch UI element to AddSoftwareCard to indicate the whether a package is closed source.
- Adds a switch UI element to SoftwareInformationForm to update whether a package is closed source.
- Updates the contents of SoftwareInformationForm responsively based on the new switch to hide the repoURL and citationField UI elements and to change the subtitle of the Keywords input.
- Updates the licenses recorded in the database for a piece of software such that when the closed source toggle is active a single entry with the value of "Proprietary" is added and when the closed source toggle is inactive licenses may be added as usual.

Feedback is sought regarding the suitability of the proposed UI and license behaviour changes.

### DIFF
--- a/frontend/components/software/add/AddSoftwareCard.tsx
+++ b/frontend/components/software/add/AddSoftwareCard.tsx
@@ -10,6 +10,8 @@ import {useEffect, useState} from 'react'
 import {useRouter} from 'next/router'
 import Button from '@mui/material/Button'
 import Alert from '@mui/material/Alert'
+import FormControlLabel from '@mui/material/FormControlLabel'
+import Switch from '@mui/material/Switch'
 import CircularProgress from '@mui/material/CircularProgress'
 
 import {useForm} from 'react-hook-form'
@@ -35,6 +37,7 @@ type AddSoftwareForm = {
   slug: string,
   brand_name: string,
   short_statement: string|null,
+  closed_source: boolean
 }
 
 let lastValidatedSlug = ''
@@ -52,7 +55,7 @@ export default function AddSoftwareCard() {
   })
   const {errors, isValid} = formState
   // watch for data change in the form
-  const [slug,brand_name,short_statement] = watch(['slug', 'brand_name', 'short_statement'])
+  const [slug,brand_name,short_statement,closed_source] = watch(['slug', 'brand_name', 'short_statement', 'closed_source'])
   // take the last slugValue
   const bouncedSlug = useDebounce(slugValue, 700)
 
@@ -142,6 +145,7 @@ export default function AddSoftwareCard() {
         brand_name: data.brand_name,
         slug: data.slug,
         short_statement: data.short_statement,
+        closed_source: data.closed_source,
         is_published: false,
         description: null,
         description_type: 'markdown',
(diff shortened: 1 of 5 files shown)


dmijatovic:
@cc-a Thank you for notifying us about the planed changes in a timely manner.  I had a look at the proposal and have few suggestions concerning the implementation points. 

- Adds a switch UI element to AddSoftwareCard to indicate the whether a package is closed source. 
**This is clear. No additional suggestions.**

- Adds a switch UI element to SoftwareInformationForm to update whether a package is closed source.
**This is clear. Maybe place the switch on the right side next to the published switch (and listen for any feedaback)?**

- Updates the contents of SoftwareInformationForm responsively based on the new switch to hide the repoURL and citationField UI elements and to change the subtitle of the Keywords input.
**The current form is somewhat complex and hiding the elements might be confusing (at least to me). Mentioned input values are not required in any case (open/closed source). I think, it is sufficient to improve the "help" messages (make these dynamic)? More complex alternative could be to disable specific input.**

- Updates the licenses recorded in the database for a piece of software such that when the closed source toggle is active a single entry with the value of "Proprietary" is added and when the closed source toggle is inactive licenses may be added as usual.
**The form currently uses "autosave" feature to patch each value individually (on lost focus). The approach is less suitable for changes of multiple values and rolling back to previous value in case of accidental change. I suggest to warn user about incorrect value (help message or snackbar) instead. This should prevent lost of "original" value(s) in case of accidental switche between open/closed source. In the specific case of licenses we need to extend or dynamically change licenses list and warn user on incorrect value(s). In addition, the software page should have additional logic and ignore specific sections or entries based on the open/closed source.** 

We have few closed source software entries in RSD currently without any "specific" requirements. @cmeessen What are your toughts about closed source software? Do you have specific "requirements" for closed source software entries? Maybe we can include this point in our weekly meetings?

I have suggestions concerning the technical implementation. The points 3 and 4 are more challenging to implement, especially within currently implemented "autosave" approach. Maybe we can set-up online meeting and see if I can offer some assistance/advice? I am available in our daily stand-ups all days except on Wednesday's.

cmeessen:
In Helmholtz we do have several use cases for closed source software, and I think this PR is already doing some important steps into the right direction :+1:. Here are my thoughts:

As far as I understand, in this implementation the major difference between an open and a closed source software is that it is automatically added a "Proprietary" license. I would therefore place the switch within the "License" block on the edit page. If a user activates the switch, the "Proprietary" license tag could be automatically added, and the license drop down as well as the repository URL field (although this could still be useful, see below) could be greyed out. This could avoid potential confusion on the users side with regard to UI consistency.

I am not sure whether the DOI can be left out entirely for a closed source package. There might still be some use cases for this nonetheless.

Here is one of our use cases of a closed source software in the Helmholtz RSD:

https://helmholtz.software/software/igmas

Note that it uses the repository URL field which points to a repo where only the binaries are available for download.

dmijatovic:
After discussion with our partners it is decided to start with simple approach by introducing a open source flag into the form and dynamically change help texts on Repository URL and Licenses. The changes will be implemented in main RSD repository in the latest RSD version (v2.x.x), see issue #1033 for the implementation details.

PR #1019, opened 2023-10-20
feat: show top 12 contributors initially

dmijatovic:
# Use  button(s) to limit initial contributors list to top 12 items

Based on the feedback from initial approach the feature is changed to:
- if there are less than 12 items, all items are shown
- if there are more than 12 items we initially show first 12 items and the `Show all` button (see image 1). Click on the button will load all items
- when more than 12 items are loaded, `Show less` button is displayed. Click on the button will reduce list to first 12 items

## Show all button
image

## Show less button
image

How to test:
* `make start` to build app
* create new software item. Use this concept DOI 10.5281/zenodo.3401363 on info page
* import contributors
* view software page. Confirm that 12 contributors and the button is shown. Load more by clicking on the button.
* reduce list by clicking on `Show less` button
* visit the software with less than 12 contributors, no buttons should be shown
* note that the position of contributor/team member can be changed in the edit section
* the same feature is available in Team section of project page

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/ContributorsList.tsx
+++ b/frontend/components/software/ContributorsList.tsx
@@ -1,46 +1,120 @@
+// SPDX-FileCopyrightText: 2022 - 2023 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-// SPDX-FileCopyrightText: 2022 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2022 dv4all
+// SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
 //
 // SPDX-License-Identifier: Apache-2.0
+
+import {useState} from 'react'
 
 import {Person} from '../../types/Contributor'
 import ContributorAvatar from './ContributorAvatar'
 import {getDisplayName, getDisplayInitials} from '../../utils/getDisplayName'
 import PersonalInfo from './PersonalInfo'
 import {getImageUrl} from '~/utils/editImage'
+import Button from '@mui/material/Button'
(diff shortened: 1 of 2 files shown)


ewan-escience (commented):
It works, but I don't think we should do this. In the example, I need to click 5 times to see all contributors, without this having any real benefit.

dmijatovic:
> It works, but I don't think we should do this. In the example, I need to click 5 times to see all contributors, without this having any real benefit.

We can discuss some variations of this approach. The idea behind is to prevent very long software page when software has a lot of contributors.

jmaassen:
Works well, but I do have some comments: 

- I agree with Ewan that with long lists of contributors, it's inconvenient that I need to click multiple times. Wouldn't it be more logical to have a "show all" instead of a "show more" ? That way only a single click is needed to expand the list. 

- At the moment there is no way to collapse the list again? I cannot get back to the "short view"

We could also opt for a more condensed view of the contributors? Although I'm not sure if this would be an improvement. Journals have similar issues with papers, and a more condensed view doesn't seem to help (famous example: https://iopscience.iop.org/article/10.1088/1748-0221/3/08/S08003#back-to-top-target)

dmijatovic:
Based on the feedback to feature is changed to:
- if there are less than 12 items, all items are shown
- if there are more than 12 items we initially show first 12 items and the `Show all` button (see image 1). Click on the button will load all items
- when more than 12 items are loaded, `Show less` button is displayed. Click on the button will reduce list to first 12 items of the list

## Show all button
image

jmaassen (approved):
Works well.

dmijatovic:
@cmeessen, @jcohen02 Do you have any objections on implementing this feature?

jcohen02 (approved):
No issue with adding this from my side, @dmijatovic. I agree with some of the comments regarding usability and needing to click multiple times, but I think this has now been rectified anyway, so I'm fine with this.

PR #1041, opened 2023-11-02
feat: enable open source / proprietary software flag

### LINKED ISSUE #1033: Add open source flag to edit software
# Open source flag for software

There is a need to have binary indication for user to indicate if the software is open source or proprietary (closed source). After discussing the feature with our partners we decided to start with basic implementation:

- Add open source / proprietary (single) choice on Add software page. Open source is default choice (image 1)
- Add same choice on Edit software page in the license section. (image 2)
   - Open source / proprietary choice
   - Find or add new license value in both cases
- Expected future adjustments:
   - For each license we can have license label and the link to a license page text. For open source licenses this info is available in spdx.json. The Concept DOI has link to license page in some cases (not always). When adding new license the user should provide url to license page.
   - We would like to refactor the content. "Software information" page could be split into "Basic info" and "Software metadata". We would like to place Concept DOI, Keywords, Licensing and maybe Contributors into "metadata" page.

## Add software 

image

## Edit software license info
image

dmijatovic:
# Add open source/proprietary software flag

Closes #1033

**Please provide feedback about current approach and the future plans.**

* Change property to open_source in software table
* Add choice Open source / Proprietary software to Add software page
* Add same choice in the license section
* Currently there are no validations on added license values.
* For the proprietary licenses the user should add "Proprietary" or other desired label using autocomplete input box
* For the open source licenses user can select values from spdx list or add the custom license label
* Section order on edit software information changed to: Status, Logo, Concept DOI, License, Keywords. Note! In the future the Logo section will moved to new Images page. We also think about creating new Software metadata page and move ConceptDOI, License and Keywords sections to that page.

How to test:
* login as user or rsd admin
* create new software. Confirm open source flag is present and pre-selected. Change to proprietary software if you wish.
* confirm the flag is persisted on the edit page in the license section
* add custom license label if you wish

## Future plans (not in this PR):
* Add url property to license_for_software table and ask the user to provide url/link to the content of the custom license. Currently user can add custom license label/name but there is no possibility to link to the license content. For the spdx open source licenses the link to license text is provided. When importing the licenses from Concept DOI, the link to license is not always provided.
* The user should be able to update/change link to license text
* Move licenses and some other section from first page to new (second) page: Software metadata section

## Example add software page
image

## Example edit software license
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/data-generation/main.js
+++ b/data-generation/main.js
@@ -121,7 +121,7 @@ async function generateSofware(amount=500) {
 			image_id: faker.helpers.maybe(() => localSoftwareLogoIds[index % localSoftwareLogoIds.length], {probability: 0.8}) ?? null,
 			is_published: !!faker.helpers.maybe(() => true, {probability: 0.8}),
 			short_statement: faker.commerce.productDescription(),
-			closed_source: !!faker.helpers.maybe(() => true, {probability: 0.8}),
+			open_source: !!faker.helpers.maybe(() => true, {probability: 0.5}),
 		});
 	}
 
--- a/database/003-create-software-table.sql
+++ b/database/003-create-software-table.sql
@@ -1,10 +1,11 @@
 -- SPDX-FileCopyrightText: 2021 - 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
--- SPDX-FileCopyrightText: 2021 - 2022 Netherlands eScience Center
+-- SPDX-FileCopyrightText: 2021 - 2023 Netherlands eScience Center
 -- SPDX-FileCopyrightText: 2022 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 -- SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 -- SPDX-FileCopyrightText: 2022 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 -- SPDX-FileCopyrightText: 2022 dv4all
 -- SPDX-FileCopyrightText: 2023 Diego Alonso Alvarez (ICL) <d.alonso-alvarez@imperial.ac.uk>
+-- SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
 -- SPDX-FileCopyrightText: 2023 Imperial College London
 --
 -- SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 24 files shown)


dmijatovic:
@cmeessen and @jcohen02 Can you have a look at this PR and provide your suggestions, please?

jcohen02:
Hi @dmijatovic, thanks for adding this functionality. I've tested it and all seems to work OK but I have a few questions/suggestions:

1) When you initially create the software, the option button for "Open Source" or "Proprietary" software is great. I think that was exactly what we had in mind.

2) However, when you then get onto the configuration page, there's nothing to specifically show that the software being edited is open source or proprietary. There is the licence section where the relevant licence type (open/proprietary) seems to be pre-selected but I get the impression there's nothing to stop you selecting software as proprietary on the first page and then assigning it an open source licence?

3) When you then view the software "cards" in the "Software" page that shows all the published software, there's also nothing there that shows you whether something is open or not - I'm not sure if this matters but I'm wondering if something like a little open or closed padlock icon at the bottom left of the card (i.e. the opposite side to the mentions icons) might be useful here - it wouldn't be editable, purely a static icon, perhaps with a mouseover pop-up that says "Open Source" or "Proprietary"?

Number 3 above is just a suggestion - I think my main concern was to make it clear that something is open source or proprietary and I think maybe this could be a bit clearer still on the software configuration page and potentially in the software list view too. 

One suggestion that was spoken about was a toggle, perhaps below the "published" toggle that enables switching between open source and proprietary - maybe that would make things clearer?

Anyway, a few suggestions, I hope these are useful and thanks for your work implementing this feature - I'm happy to provide more detail/feedback if useful.

dmijatovic:
@jcohen02 Thanks for the feedback. 

My apology for a late reaction, I needed some time to reflect on implementation of this feature. I was not completely satisfied with the current solution. After talking to other members of the RSD team we came to conclusion that property open/closed actually belongs to license (type) and not directly to the software. We received feedback about some edge cases which use multiple licenses with the possible mix of open/closed source licenses for some parts of the software. The value, when aggregated at the software level, could be potentially `open source`, `proprietary` and `mixed open/proprietary`. Currently I favor the approach where this value is deducted by RSD from specified software license and shown per license. In other words, the user should provide the license(s) information only. The RSD should deduct the license type for the most common licenses and only for the custom types, which user can create manually, we ask user to provide open/closed source flag and (optionally) the link/url to "complete" license text. 

We only record license name at the moment which is insufficient. I propose to extend the license information to name, type and the url. When adding a new license we ask user to provide required additional information. Once added to RSD a new license could be used by others too (it will appear in the list of licenses). 

I believe this approach records the information at the optimal level (license) and give us flexibility to support any type of license. On the software page we can list the licenses including a link to license text (if provided). We could also use open/closed icon to indicate the type of the license.

I am interested to hear what others think about this approach. @cmeessen, @ctwhome, @fembau, @jmaassen, @ewan-escience

jcohen02:
Thanks for the feedback and your further thoughts on this @dmijatovic.

I see your point in terms of using the licence type to deduce whether the code - I guess this serves the purpose of being able to set whether a piece of software is open or closed source and to be able to search based on that. At the same time, having a simple "switch" as you had in the original implementation to say "open source" or "not open source" feels like it has as a greater element of simplicity to it. However, I appreciate that this could also then end up with someone stating that a piece of software is open source and then specifying a licence that is proprietary. In that context, your suggestion makes more sense.

It would be useful to hear others' thoughts on this. I'm thinking that your suggested route forward makes most sense so happy from my side if you decide to go forward with that approach.

PR #1058, opened 2023-11-24
Admin role in database

ewan-escience:
## Save admins in database

* RSD admins are now defined in a new table called `admin_account`.
* Consequently, the env variable `RSD_ADMIN_EMAIL_LIST` is dropped.

How to test:

* Log in in your preferred way. You should not have access to any admin features. Go to your settings to grab your account ID.
* Connect to the database and execute the following query: `INSERT INTO admin_account VALUES ('00000000-0000-0000-0000-000000000000');`, where you replace the UUID with the value you got in the previous step.
* Log out and log in with the same account. You should now have access to all admin features.
* Log in with a different account. You should not have admin access.
* Repeatedly logging in and out with different accounts should work.

PR Checklist:

*   [x] Update documentation
*   [ ] Tests


authentication/src/main/java/nl/esciencecenter/rsd/authentication/PostgrestAccount.java
```diff
@@ -31,7 +31,7 @@ public AccountInfo account(OpenIdInfo openIdInfo, OpenidProvider provider) throw
 		String subject = openIdInfo.sub();
 		String subjectUrlEncoded = URLEncoder.encode(subject, StandardCharsets.UTF_8);
 		String providerUrlEncoded = URLEncoder.encode(provider.toString(), StandardCharsets.UTF_8);
-		URI queryUri = URI.create(backendUri + "/login_for_account?select=id,account,name&sub=eq." + subjectUrlEncoded + "&provider=eq." + providerUrlEncoded);
+		URI queryUri = URI.create(backendUri + "/login_for_account?select=id,account_id:account,name,account(admin_account(account_id))&sub=eq." + subjectUrlEncoded + "&provider=eq." + providerUrlEncoded);
```
jcohen02, line 34: Minor point but it's perhaps a little unclear what is happening here in terms of the URI being created and its purpose. As a result, the reason for the changes here feels a little challenging to understand. Maybe this would benefit from a comment with some description. e.g. this is, as I understand it, formulating the login URL that is used alongside the generated token for managing logins - the database record now contains a boolean to state whether a user is an admin or not so this is being added to the URL? Where was this admin information previously being included when it was not coming from the database? Apologies if I'm completely misunderstanding.

jcohen02 (commented):
This looks good @ewan-escience, there's one bit that I'm a little unclear about that I think could possibly benefit from a comment explaining what's going on.

authentication/src/main/java/nl/esciencecenter/rsd/authentication/PostgrestAccount.java
jcohen02: 👍 The updates look good.

ewan-escience: Thanks for your feedback. I just added some comments that hopefully clear things up about the URI.

Admins were first defined through the value of `RSD_ADMIN_EMAIL_LIST` in your `.env` file. This value would contain a list of email addresses of admins. We would check if the email address that we got back from the authentication provider was in the list and if yes, admin privileges would be given.

This system is based on that we can trust that the authentication providers check that the email address that the user provides is indeed owned by that user. We think that's a potential security issue and therefore decided to switch to this system.

jcohen02: That all makes sense, thanks for the clarification.

jcohen02 (approved):
Following the update to the documentation, this looks good to me.

jmaassen (approved):
Works as expected. 

We should make sure this is easy to do if someone is setting up a remote RSD server somewhere. For example, provide some script or command line that helps doing the insert.

cmeessen (approved):

PR #1066, opened 2023-12-04
rsd-admin can add/remove rsd-admin role to other users

dmijatovic:
# Assign rsd admin role

*  RSD admin can assign/remove rsd-role to other rsd users using rsd users page in the administration section    
* Documentation page of rsd users is updated with new animation

How to test:
* `make start` to build solution and create test data
* login via SURFconext, as professor1 for example
* make professor1 user as rsd-admin via database update (see documentation)
* login via SURFconext as professor3 for example in order to create rsd account for the second user
* login as professor1 again and navigate to adminstration/rsd-users page
* search for professor3 by name `Newton` 
* assign professor3 role of rsd-admin by changing the switch near the name
* login as professor3 and confirm that it has rsd-admin role

## Change professor3 to rsd-admin

image

## Cannot change its own role
image

PR Checklist:

*   [x] Update documentation
*   [ ] Tests


jmaassen (approved):
Works as expected. 

I did notice one race condition. If two users, both admin, are logged in at the same time in a different browser, they can delete each others admin rights and that way delete the last admin ;-)

Chances of this happening are pretty much zero, so not a real issue I think

ewan-escience:
> I did notice one race condition. If two users, both admin, are logged in at the same time in a different browser, they can delete each others admin rights and that way delete the last admin ;-)

You can then an admin directly in the database again, so I'm ok with this.

PR #1068, opened 2023-12-06
Easily become admin when developing locally

### LINKED ISSUE #1062: Add support for dev enironment in the authentication service
We have changed the way the rsd admin role is assigned to users. Since version v 2.4.0 the account id of rsd admin user need to be defined in the admin_account table. This is done as security improvement measure.

During the development it is more suitable to automatically add first person to have rsd_admin role. 
I would like to propose the following approach for development mode:

- Add `RSD_ENV` variable to authentication service
- When `RSD_ENV=development` is defined in the .env file all of the authentication services (at lease SURCconext and Local) should assign rsd_admin role to first user logged in (if admin_account is empty add users account id).

@cmeessen, @fembau and @jcohen02 Do you agree to this proposal or maybe you have some other suggestion how to make rsd_admin role easily accessible during the development?

ewan-escience:
## Easily become admin when developing locally

* Add a new environment variable `RSD_ENVIRONMENT` that when set to the value `dev`, allows the first user that logs in to become an admin. With any other value, this env variable does nothing.

**Note:** This is a feature for local development only and is therefore not safe in production. When two people login at the same time, they might both become admin for example.

How to test:

* Add the new env variable to your `.env` and set its value to `dev`.
* Log in in any way you like, you should be an admin. Repeatedly logging in and out with this account should not change this. Repeatedly logging in and out with other accounts should *not* make those admin.
* Run `DELETE FROM admin_account;` in the database to delete all existing admins.
* When you now log in with any account, that account should become admin.
* Delete all admins again, change the value of `RSD_ENVIRONMENT` to anything other than `dev` and restart the RSD (or at least the auth service with `docker compose up auth`)
* Logging in should *not* make you admin.

Closes #1062

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,11 @@ POSTGREST_URL_EXTERNAL=http://localhost/api/v1
 # .env.local: http://localhost/auth, .env: http://auth:7000
 RSD_AUTH_URL=http://auth:7000
 
+# consumed by services: authentication
+# If set to "dev", the first user to log in will become admin.
+# Any other value doesn't activate this feature (and doesn't do anything).
+RSD_ENVIRONMENT=prod
+
 # consumed by services: authentication, frontend (api/fe)
 # provide a list of supported OpenID auth providers
 # the values should be separated by semicolon (;)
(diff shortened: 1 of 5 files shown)


dmijatovic (approved):
Nice work!

PR #1090, opened 2024-01-23
Don't allow user to set updated_at columns for account

ewan-escience:
## Don't allow user to set updated_at columns for account

* Silently reset the `agree_terms_updated_at` and `notice_privacy_statement_updated_at` columns when a user changes this in the `account` table.
* Add a test for this and refactor some older tests
* Update dependencies of the backend tests
* Add a line to the docs explaining to add the setup class

How to test:

* Inspect the test to see if it does what it is supposed to
* Build and run the tests: `make run-backend-tests`

I'm not entirely sure if silently overwriting is better than generating an error. If someone has a good argument for the latter, I might change it.

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [x] Tests

### DIFF
--- a/backend-tests/README.md
+++ b/backend-tests/README.md
@@ -1,6 +1,6 @@
 <!--
-SPDX-FileCopyrightText: 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-SPDX-FileCopyrightText: 2023 Netherlands eScience Center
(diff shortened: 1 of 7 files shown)


dmijatovic (approved):
This works as far as I can see. I am not expert in Java tests so someone else should take a look at this part.

One question though, are these tests part of the CI pipeline?

Issue #1097, opened 2024-01-26
Disable port mapping in frontend-dev


cmeessen:
The port mapping `3000:3000` in `docker-compose.dev.yml` is not required because this is already being taken care of in nginx.

PR #1100, opened 2024-01-26
1095 distinguish openid providers

### LINKED ISSUE #1095: Distinguish between OpenID providers used for login and for account coupling
Public profile pages require ORCID to be set up via the environment variables. This implies that, when ORCID is enabled, login is also always possible using ORCID.

For the Helmholtz instance we would like to allow coupling with ORCID but not allow login via ORCID.

cmeessen:
# Distinguish between login and coupling auth providers

Fixes #1095 #1097

* Distinguish between auth providers used for login and for account coupling. This allows couple an ORCID account to an RSD account without having to allow ORCID login simultaneously
* :exclamation: This PR adds a new environment variable `RSD_AUTH_COUPLE_PROVIDERS`
* contains minor fix: removed the port mapping for the frontend dev mode because this was not necessary

How to test:

* in `.env`, set:
  ```
  RSD_AUTH_PROVIDERS="SURFCONEXT;LOCAL"
  RSD_AUTH_COUPLE_PROVIDERS=ORCID
  ```
* `docker compose build --parallel && docker compose up --scale scrapers=0`
* open login dialogue and verify that providers are Surf and Local only
* login
* navigate to user profile and link an ORCID account to the rsd account
* `docker compose down`
* add `ORCID` to `RSD_AUTH_PROVIDERS` and repeat previous steps. Verify that ORCID login works
* `docker compose down`
* in `.env`, comment out `RSD_AUTH_COUPLE_PROVIDERS`, or set it to something else than `ORCID`
* launch RSD and verify that ORCID login works but coupling button is not shown
* open the user documentation http://localhost/documentation/rsd-instance/configurations/ and check for updated content about the ORCID integration

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,6 +73,12 @@ RSD_ENVIRONMENT=prod
 # if you add the value "LOCAL", then local accounts are enabled, USE THIS FOR TESTING PURPOSES ONLY
 RSD_AUTH_PROVIDERS=SURFCONEXT;ORCID;AZURE;LOCAL
 
+# consumed by services: authentication, frontend (api/fe)
+# provide a list of supported OpenID auth providers for coupling with the user's RSD account
+# the values should be separated by semicolon (;)
+# Allowed values are: ORCID
+RSD_AUTH_COUPLE_PROVIDERS=ORCID
+
 # Define a semicolon-separated list of user email addresses which are allowed to
 # login to the RSD. If the variable is left empty, or is not defined, all users
 # will be allowed to login.
--- a/authentication/src/main/java/nl/esciencecenter/rsd/authentication/Config.java
+++ b/authentication/src/main/java/nl/esciencecenter/rsd/authentication/Config.java
@@ -1,10 +1,10 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-// SPDX-FileCopyrightText: 2022 - 2023 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 // SPDX-FileCopyrightText: 2022 - 2023 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2022 - 2024 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 Matthias Rüster (GFZ) <matthias.ruester@gfz-potsdam.de>
 // SPDX-FileCopyrightText: 2022 dv4all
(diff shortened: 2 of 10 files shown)


cmeessen:
Waiting for #1103 to be closed by @ewan-escience

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Config.java
```diff
@@ -22,7 +22,7 @@ public static String jwtSigningSecret() {
 	}
 
 	private static Collection<String> rsdAuthProviders() {
```
ewan-escience, line 24: I would rename this method to `rsdAuthCoupleProviders`

ewan-escience (changes requested):
Looks good, just one small renaming needed.

ewan-escience:
It is also good to note that when ORCID login is disabled and ORCID coupling is enabled, when an ORCID is coupled, this ORCID is still added to the login allow list.

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Config.java
cmeessen: Good point, I updated this.

cmeessen:
> It is also good to note that when ORCID login is disabled and ORCID coupling is enabled, when an ORCID is coupled, this ORCID is still added to the login allow list.

I added a warning about this into the documentation, and also split it up into sections for common variables, coupling and authentication respectively. Could you please double check whether all common variables are mentioned correctly?

ewan-escience:
> Could you please double check whether all common variables are mentioned correctly?

Everything seems to be there. You used as example`ORCID_REDIRECT_COUPLE=http://YOUR-RSD-DOMAIN/auth/couple/orcid`, maybe you can also use `YOUR-RSD-DOMAIN` for `ORCID_REDIRECT`.

.env.example
```diff
@@ -71,7 +71,13 @@ RSD_ENVIRONMENT=prod
 # Allowed values are: SURFCONEXT, HELMHOLTZAAI, ORCID or LOCAL
 # if env value is not provided default provider is set to be SURFCONEXT
 # if you add the value "LOCAL", then local accounts are enabled, USE THIS FOR TESTING PURPOSES ONLY
-RSD_AUTH_PROVIDERS=SURFCONEXT;ORCID;AZURE;LOCAL
+RSD_AUTH_PROVIDERS="SURFCONEXT;ORCID;AZURE;LOCAL"
```
dmijatovic, line 74: Do we need `"`? I think it works with or without `"` and we do not use in other variables.

cmeessen: I think I added it because I had an issue with the `;` being interpreted as a command separator. But I couldn't reproduce it and removed the `"`.

ewan-escience (approved):

PR #1101, opened 2024-01-26
Docs setup instance

ewan-escience:
This PR cleans up the previous external PR.


documentation/docs/03-rsd-instance/01-getting-started.md
```diff
@@ -79,3 +79,51 @@ When you log in to the RSD as administrator, you will see an additional "Adminis
...
+
+You will need git:
+
+```env
+sudo apt install git
+```
+
+You will also need Docker with the Compose plugin, see the installation instructions for [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/).
+
+Then you can then clone the RSD:
```
dmijatovic, line 96: Then is used twice

dmijatovic (approved):
Looks good!

PR #1105, opened 2024-01-30
fix: only call ORCID well known endpoint on settings page

### LINKED ISSUE #1103: Unnecessary requests to ORCID
On all user pages (e.g. `/user/software`, `/user/projects`), a request is made to the ORCID well known endpoint from the Next.js server. This causes navigation to these pages to be slow (as these requests usually take ~500ms) and might cause 404 pages if the request fails for some reason.

ewan-escience:
## Fix ORCID requests

* Only call ORCID well known endpoint on section page
* Replace string `'MISSING'` with `null` if coupling URL is missing
* Wrap code in try-catch statement to always be able to show the settings page

How to test:

* Log in with a non-ORCID account
* Navigate various user pages, like http://localhost/user/projects and http://localhost/user/settings. There should only be a noticable delay on the settings page

**Note:** It would be better to only fetch the data when needed when clicking the button, or asynchronously in the background. It might also be good to use caching. However, this PR only implements the minimum necessary, as this code is also being worked on in #1100.

Closes #1103

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/auth/api/authHelpers.ts
+++ b/frontend/auth/api/authHelpers.ts
@@ -1,7 +1,8 @@
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 dv4all
+// SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2023 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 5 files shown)


cmeessen (approved):

dmijatovic (approved):
I have created new method to handle ORCID auth info and keep it in memory. It will refresh orcid_endpoint info per hour.

To test: enable logs in `apiLinkOrcidProps` and observe when request are made.

cmeessen (approved):
Just for the protocol: using this implementation, the LinkOrcid button only is shown if `ORCID_REDIRECT_COUPLE` is defined.

dmijatovic:
> Just for the protocol: using this implementation, the LinkOrcid button only is shown if `ORCID_REDIRECT_COUPLE` is defined.

That is correct. We can apply additional logic in the apiLinkOrcidProps methods if needed.

ewan-escience:
It is good to write down here that this PR now also contains caching.

PR #1111, opened 2024-02-06
Improve global search function

### LINKED ISSUE #1108: No match when searching exact title
In some cases, the search function does not return the correct results even though the user enters the exact software title.

Example from [helmholtz.software](https://helmholtz.software):

image

This is the software entry that should appear: https://helmholtz.software/software/rce

ewan-escience:
## Improve global search function

* Use SQL instead of PL/pgSQL as the language for the `global_search` function
* Only search on slug and title (not keywords, research domains or short descriptions anymore)
* Give each result a score, based on exact match on the slug or title, the query being a prefix of the slug or title, or the query appearing anywhere in the slug or title
* Adapt the frontend to use the new function

How to test:

* Test out the global search with the fake data
* Log in and create software pages with the following titles: `abc`, `abc def` and `ghi abc def`. Global search on `abc`, they should be returned in the order above.

**Discussion:** Should the function enforce a minimum length for its input?

Closes #1108

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/database/100-create-api-views.sql
+++ b/database/100-create-api-views.sql
@@ -1098,7 +1098,8 @@ CREATE FUNCTION global_search(query VARCHAR) RETURNS TABLE(
 	name VARCHAR,
 	source TEXT,
 	is_published BOOLEAN,
-	score INTEGER
+	rank INTEGER,
+	index_found INTEGER
 ) LANGUAGE sql STABLE AS
 $$
 	-- SOFTWARE search item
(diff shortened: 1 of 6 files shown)


cmeessen:
I can confirm that it's working as described :+1:

Additionally, I checked how it is performing with the generated data searching for "duh". In my case I get the following results:

image

It does put the project match before organisation, although the organisation starts with "duh" whereas the project has the term on the second place.

> Discussion: Should the function enforce a minimum length for its input?

I think searching from the first letter may not make much sense. But two letters could already yield some meaningful results.

> Only search on slug and title (not keywords, research domains or short descriptions anymore)

Searching in short descriptions was a request in the past, if I remember correctly. Would it be possible to include this again?

ewan-escience:
> It does put the project match before organisation, although the organisation starts with "duh" whereas the project has the term on the second place.

Note that `Project: ` and `Organisation: ` are part of their names, so it currently works as intended, since we don't look at the index of the query. We could do this, e.g. use the index as a tie breaker when sorting. Would you like that?

> I think searching from the first letter may not make much sense. But two letters could already yield some meaningful results.

Agreed. However, SQL doesn't have an `IF` statement. So we could either do a complicated `CASE` statement, revert back to PL/pgSQL, only do that on the frontend, or leave it as it is.

> Searching in short descriptions was a request in the past, if I remember correctly. Would it be possible to include this again?

Personally, I feel like this doesn't belong on a global search and that one can use the dedicated software search page for this. However, if more people think this is useful, I could add it back. I still have to think about how that should interact with the scores though.

dmijatovic:
I am in favor of including subtitle/short description in the search.

jmaassen:
Works well, but I also wonder if it wouldn't be better to also search in the short description. 

To me, the global search is a quick way to search though all three collections (software, projects, organisations). Often, I don't remember the exact name of software or project, but known it had something to do with a specific topic. A good example is "fusion" as a global search, which gives me things like [duqtools](https://research-software-directory.org/software/duqtools) and [imas2xarray](https://research-software-directory.org/software/imas2xarray).

ewan-escience:
I updated the `global_search` function to include keywords for software and projects; short descriptions for software; subtitles for projects.

When searching in the frontend, the results are returned in the following order, keeping in mind that all matches are case insensitive:

1. software, projects and organisations with an exact match on title, name or slug
2. software and projects where there is an exact match on one of their keywords
3. software, projects and organisations where the query is a prefix of their title, name or slug
4. software, projects and organisations where the query appears anywhere in their title, name or slug; these are then ordered by the index in which the match appears
5. software, projects and organisations when the query is anywhere in the short description, subtitle or part of any keyword

Please test this with your own test cases and inspect the SQL code. Note that we can't satisfy all use cases and that we can refine it later.

Also test this with a larger data set. I expect it's fine for now (I tested this on production data), but as we grow, it might be too slow, especially when searching the keywords. We might need dedicated search engines then.

I also adapted the frontend so that it only searches when at least two character are given.

dmijatovic (approved):
It works well for me. Let's see what others are thinking.

Issue #1081, opened 2024-01-09
Add "support" for additional package managers


dmijatovic:
Based on current RSD data we should extend the package manager list with:

- Debian: https://packages.debian.org/search?keywords=libcdk-java
- Sonatype: https://central.sonatype.com/artifact/org.openscience.cdk/cdk-bundle/overview
- Chocolatey: https://community.chocolatey.org/packages/openchrom
- Snapcraft: https://snapcraft.io/openchrom
- Github: https://github.com/orgs/research-software-directory/packages?repo_name=RSD-as-a-service 
- Gitlab: https://gitlab.com/gitlab-org/gitlab/container_registry
- Rust: https://crates.io/crates/serde
- Go: https://pkg.go.dev/strconv

The ENUM in db should be extended with additional keys and the logos should be obtained from official sources.
Below is complete (new) list of package managers we "recognize" and the scraper services we support. If we do not have scraper service the services array is empty (no scrapers running).

**In addition, package manager data should be updated from 'other' to proper enum value after the upgrade**

## Updated package manager configuration object on FE

```Javascript
export const packageManagerSettings = {
  anaconda: {
    name: 'Anaconda',
    icon: '/images/anaconda-logo-96.svg',
    hostname: ['anaconda.org'],
    services: ['dependents']
  },
  cran: {
    name: 'CRAN',
    icon: '/images/cran-r-logo.svg',
    hostname: ['cran.r-project.org'],
    services: ['dependents']
  },
  chocolatey: {
    name: 'Chocolatey',
    icon: '/images/chocolatey-logo.svg',
    hostname: ['community.chocolatey.org'],
    services: []
  },
  debian:{
    name: 'Debian',
    icon: '/images/debian-logo.svg',
    hostname: ['packages.debian.org'],
    services: []
  },
  dockerhub: {
    name: 'Dockerhub',
    icon: '/images/dockerhub-logo.webp',
    hostname: ['hub.docker.com'],
    services: ['downloads']
  },
  github: {
    name: 'Github',
    icon: '/images/github-logo.svg',
    hostname: ['github.com'],
    services: []
  },
  gitlab: {
    name: 'Gitlab',
    icon: '/images/gitlab-icon-rgb.svg',
    hostname: ['gitlab.com','registry.gitlab.com'],
    services: []
  },
  maven: {
    name: 'Maven',
    icon: '/images/apache-maven-logo.svg',
    hostname: ['mvnrepository.com'],
[...]

PR #1116, opened 2024-02-13
974 improve maven scraper

### LINKED ISSUE #974: Improve Maven package manager
# Feedback from user

I tried the following:
 
* Link to Maven Central  https://repo.maven.apache.org/maven2/org/key-project/key.core/2.12.0/
* Maven Coordinates: "org.key-project:key.core:2.12.0"
* The PackageURL: pkg:maven/org.key-project/key.core@2.12.0
* Other URLs:
   * https://mvnrepository.com/artifact/org.key-project/key.core/2.12.0
   * https://central.sonatype.com/artifact/org.key-project/key.core/2.12.0

I just have looked into the source code of RSD, and found the following pattern:

```java
private static final Pattern urlPattern = Pattern.compile("[https://mvnrepository\\.com/artifact/([^/]+)/([^/]+)/](https://mvnrepository//.com/artifact/([%5E/]+)/([%5E/]+)/)?");
```

The following url worked: https://mvnrepository.com/artifact/org.key-project/key.core
You may want to add a documentation entry, because [mvnrepository.com](http://mvnrepository.com/) is also not the official source.

ewan-escience:
## More package manager scrapers

* Add package manager scrapers for Sonatype, Go and Crates.io.
* Fix small bug where `0` is reported when `null` is returned from the database for package managers
* Improve package manager examples to data generation
* Move scrapers to dedicated package

How to test:

* Create a software page with the following package managers:
	* https://central.sonatype.com/artifact/org.openscience.cdk/cdk-bundle/
	* https://crates.io/crates/tokio
	* https://pkg.go.dev/github.com/gin-gonic/gin
	* https://pkg.go.dev/google.golang.org/grpc
* Run the scraper: `docker compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.package_manager.MainPackageManager`
* Check the results on the Package manager tab and the Background services tab
* Login as admin and check the error logs, they should have no related errors
* Also add https://pkg.go.dev/google.golang.org/does/not/exist and run the scaper again; the Package manager should report `no info` instead of `0`

Closes #974
Related to #1081

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [x] Tests

### DIFF
--- a/data-generation/real-data.js
+++ b/data-generation/real-data.js
@@ -83,13 +83,14 @@ const packageManagerLinks = [
 	{url: 'https://pypi.org/project/cffconvert/', type: 'pypi'},
 	{url: 'https://pypi.org/project/flask', type: 'pypi'},
 	{url: 'https://packages.debian.org/search?keywords=libcdk-java', type: 'debian'},
-	{url: 'https://central.sonatype.com/artifact/org.openscience.cdk/cdk-bundle/overview', type: 'sonatype'},
+	{url: 'https://central.sonatype.com/artifact/org.openscience.cdk/cdk-bundle/', type: 'sonatype'},
 	{url: 'https://community.chocolatey.org/packages/openchrom', type: 'chocolatey'},
 	{url: 'https://snapcraft.io/openchrom', type: 'snapcraft'},
 	{url: 'https://github.com/orgs/research-software-directory/packages?repo_name=RSD-as-a-service', type: 'github'},
 	{url: 'https://gitlab.com/gitlab-org/gitlab/container_registry', type: 'gitlab'},
-	{url: 'https://crates.io/crates/serde', type: 'crates'},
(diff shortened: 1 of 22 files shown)


dmijatovic (approved):
Nice work!

jmaassen (approved):
Works well.

jmaassen:
Small question: looking at the code, it seems the frontend needs to know which scrapers there are and what data they produce. It has a has coded list of scrapers and fields like dependents and/or downloads. Wouldn't it be easier in the long run if this solution was more modular and allowed the backend to simply provide a list to the frontend of scraper configs?

ewan-escience:
We could co that, but it would require us to setup some server with this info. If it becomes a problem later, we can have another look.

PR #1124, opened 2024-02-16
Upgrade download and upload actions

### LINKED ISSUE #1123: Upgrade upload and download actions
In #1119, I upgraded `actions/download-artifact` from v3 to v4. However, when running the release action today, this resulted in errors. This might be because I forgot to upgrade `actions/upload-artifact` from v3 to v4. We temporarilly solved it by downgrading `actions/download-artifact` back to v3, but we should look into upgrading both.

ewan-escience:
Closes #1123


dmijatovic (approved):
Excellent!

Issue #1117, opened 2024-02-13
Allow admins to disable scraping


ewan-escience:
Some things being scraped will always fail, e.g. a non-public GitLab repo. These will, currently, keep on being scraped forever, generating errors and delaying other things to be scraped. It would be nice if an admin can disable certain things to be scraped (based on the error logs).

PR #1153, opened 2024-03-28
1117 disable scraping

ewan-escience:
## Disallow Git scraping

* Admins can disable the scraping of a Git repo, maintainers can see the reason why

How to test:

* Sign in as admin, create two software pages, set a repo for both, then fill the disallow scraping reason in for one of them
* Wait for the scrapers to run
* After a few minutes, the page without a reason should be scraped, the other one should not
* On the background services tab, you should be able to see a reason, both as admin and maintainer
* On the links & metadata tab, the edit reason field should only be visible for admins

Related to #1117 (we still need to do this for package managers)

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/components/software/edit/links/AutosaveRepositoryUrl.tsx
+++ b/frontend/components/software/edit/links/AutosaveRepositoryUrl.tsx
@@ -1,10 +1,10 @@
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2023 - 2024 Dusan Mijatovic (Netherlands eScience Center)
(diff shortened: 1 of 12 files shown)


frontend/components/software/edit/links/AutosaveRepositoryUrl.tsx
```diff
@@ -128,6 +128,17 @@ export default function AutosaveRepositoryUrl() {
     }
   }, [urlError, repository_url, platform.id])
 
+  async function saveScrapingDisabledReason({value}: any) {
+    return fetch(`/api/v1/repository_url?software=eq.${id}`, {
```
dmijatovic, line 132: Can you please handle the error in try/catch and show it to using showErrorMessage. The value should not be any but a string|null, right?

dmijatovic (approved):
Nice job!

PR #1161, opened 2024-04-09
fix: update contact people when new contact person is set

### LINKED ISSUE #948: It is possible to make contact person to more than one member in projects

ewan-escience:
## Prevent multiple contact people

* When a contributor or team member is made to be a contact person, the other contributors/team members are set *not* to be a contact person

How to test:

* Sign in as admin, edit a software contributor to be a contact person, the other contributors should not be a contact person anymore
* Repeat when adding a new contributor
* Try the same for the team members of a project

Closes #948

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/components/projects/edit/team/index.tsx
+++ b/frontend/components/projects/edit/team/index.tsx
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
-// SPDX-FileCopyrightText: 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-// SPDX-FileCopyrightText: 2022 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2022 - 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2022 - 2024 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2023 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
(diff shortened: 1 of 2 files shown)


frontend/components/projects/edit/team/index.tsx
```diff
@@ -75,17 +75,33 @@ export default function ProjectTeam() {
...
+      updateContactPeople(member, newMembersList)
       setMembers(newMembersList)
     }
   }
 
+  async function updateContactPeople(data: TeamMember, list: TeamMember[]) {
+    if (!data.is_contact_person) return
+
+    list.forEach(person => {
+      if (!person.is_contact_person || person === data) return
```
dmijatovic, line 95: Does person === data works properly? In JS this compares the reference (ids) of these object and not the content, so it never be equal ASFAIK.

frontend/components/projects/edit/team/index.tsx
```diff
@@ -75,17 +75,33 @@ export default function ProjectTeam() {
...
       const newMembersList = [
         ...members,
         member
       ]
+      updateContactPeople(member, newMembersList)
       setMembers(newMembersList)
     }
   }
 
+  async function updateContactPeople(data: TeamMember, list: TeamMember[]) {
```
dmijatovic, line 91: I would prefer if the api call function is in the same file where other api calls of team section are.

frontend/components/projects/edit/team/index.tsx
ewan-escience: Yes, it works, because data was added to the list just before in the calling function.

dmijatovic (approved):
Interesting approach :-). I thinks we can close #1004 too with this (for now)

PR #1165, opened 2024-04-12
ci: add token for release workflow

ewan-escience:
This should allow the release workflow to push the `CITATION.cff` file without requiring us to disable branch protection. Followed instructions from [here](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#push-to-protected-branches).

### DIFF
--- a/.github/workflows/_cff.yml
+++ b/.github/workflows/_cff.yml
@@ -33,6 +33,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{inputs.branch}}
+          token: ${{ secrets.PAT_RELEASE }}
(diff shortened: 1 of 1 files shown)


.github/workflows/_cff.yml
```diff
@@ -45,3 +46,4 @@ jobs:
         with:
           commit_message: ${{inputs.commit_message}}
           file_pattern: CITATION.cff
+          push_options: --force-with-lease
```
dmijatovic, line 49: I would prefer not to force push this commit.

dmijatovic (approved):

PR #1180, opened 2024-04-24
1179 ror requests

### LINKED ISSUE #1179: Requests to ROR made when ROR ID is missing
When an organisation doesn't have a ROR ID, the Next.js server still makes requests to the ROR API (requesting data from https://api.ror.org/organizations/null).

ewan-escience:
## Remove unnecessary requests to ROR API

* Prevent the Next.js server from making requests to the ROR API when the ROR ID is missing on organisation pages
* Fix some typos in the frontend README
* Add a section to the frontend README explaining how to intercept HTTP requests on the Next.js server

How to test:

* Start the frontend dev server with `yarn dev`
* Follow the new instructions on the README
* On http://localhost:3000, check out an organisation without a ROR ID, no requests should be made to the ROR API
* Compare to an organisation that has a ROR ID

Closes #1179

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/utils/getROR.test.ts
+++ b/frontend/utils/getROR.test.ts
@@ -1,0 +1,86 @@
+// SPDX-FileCopyrightText: 2024 Dusan Mijatovic (dv4all) (dv4all)
+// SPDX-FileCopyrightText: 2024 dv4all
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import {getOrganisationMetadata, findInROR} from './getROR'
+
+// mock fetch
+const mockFetchJson = jest.fn((props) => Promise.resolve(props))
+const mockFetch = jest.fn((props) => {
+  return Promise.resolve({
+    status: 200,
+    ok: true,
+    json: ()=>mockFetchJson(props)
+  })
+})
+global.fetch = mockFetch as any
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
+
+it('getOrganisationMetadata does NOT call fetch on undefined ror_id', async () => {
+  const ror_id = undefined
+  const resp = await getOrganisationMetadata(ror_id as any)
+  expect(resp).toBe(null)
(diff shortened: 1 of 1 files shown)


frontend/utils/getROR.ts
```diff
@@ -68,8 +70,7 @@ function buildAutocompleteOptions(rorItems: RORItem[]): AutocompleteOption<Searc
 export async function getOrganisationMetadata(ror_id: string|null) {
   try {
     // check availability
-    if (typeof ror_id === 'undefined') return null
-    if (ror_id === null && ror_id === '') return null
+    if (ror_id === undefined || ror_id === null || ror_id.trim().length === 0) return null
```
dmijatovic, line 73: Does this work properly? I learned to use typof ror_id == "undefined"

ewan-escience: Yes, see e.g. [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#strict_equality_and_undefined).

dmijatovic: Great link, the section about [using typeof operator on that page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined) explains the difference between the approaches. I prefer the approach that does not throw an error. In this particular case the condition `ror_id === undefined` can be removed?

ewan-escience: It will never throw a `ReferenceError`, because `ror_id` is the function parameter and will therefore always exist as reference. And otherwise you would also have to do that for every other variable ever, cause the error will also be thrown on `null` and all other values.

dmijatovic (approved):

cmeessen (approved):

PR #1193, opened 2024-05-06
fix: only extract mention id if present when not able to update a men…

ewan-escience:
Some mentions have an external OpenAlex ID but no DOI in our database, even though they do have a DOI in OpenAlex. This PR fixes a wrong case of exception handling in that case.


dmijatovic (approved):

jmaassen (approved):
Seems okay.

PR #1201, opened 2024-05-15
store message with metadata in the error logs when saving a mention fails

ewan-escience:
Store the DOI, external ID and source in the error logs when a mention couldn't be saved in the database for whatever reason.


jmaassen (approved):
Looks good.

Issue #1198, opened 2024-05-13
Upgrade eslint naar v9 and setup prettier


dmijatovic:
We should upgrade eslint to v9 and use prettier as formatter. 

Currently we use eslint for both linting and formatting. 
The formatting rules will be [removed from eslint in the future](https://eslint.org/blog/2023/10/deprecating-formatting-rules/). 

We will need to update eslint and configure prettier.

ewan-escience:
A migration guide can be found [here](https://eslint.org/docs/latest/use/migrate-to-9.0.0). Notable is that the configuration format has changed, for which a [migration quide](https://eslint.org/docs/latest/use/configure/migration-guide) exists as well.

ewan-escience:
Since 14 October, the [issue for Eslint 9 and Next.js](https://github.com/vercel/next.js/issues/64409) has been resolved, so we could have another try,

PR #1203, opened 2024-05-16
Prettier for data generation

ewan-escience:
## Prettier for data generation

* Configure and apply [Prettier](https://prettier.io/) to the data-generation module
* Add a GitHub workflow that fails when this module is not properly formatted with Prettier
* Add a readme for the data-generation-module

How to test:

* This should run without errors
* Check the result(s) of the workflow(s)
* Check the file changes
* Check the new `README.md`

Related to #1198
Related to #915

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/.github/workflows/prettier-check-data-generation.yml
+++ b/.github/workflows/prettier-check-data-generation.yml
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+# SPDX-FileCopyrightText: 2024 Netherlands eScience Center
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: "Run Prettier check"
+on:
+  pull_request:
+    paths:
+      - "data-generation/**"
+  workflow_dispatch:
+jobs:
+  run-prettier-check:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: "data-generation"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: "npm"
+          cache-dependency-path: "data-generation/package-lock.json"
+      - run: npm clean-install
+      - run: npm run format:check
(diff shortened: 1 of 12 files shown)


.github/workflows/prettier-check-data-generation.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: "data-generation"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+      - run: npm install --save-exact prettier
```
dmijatovic, line 23: I have few suggestions
- Please include prettier in devDependencies in the package json. `npm i -D prettier`
- Please create npm scripts for checking and fixing, something like `format:check` and `format:fix`
- Use these values in the action: 
   - npm install
   - npm run format:check

Extend readme with explanation how to locally run check and fix formatting.

data-generation/.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+	"printWidth": 120,
+	"useTabs": true,
+	"tabWidth": 4,
+	"semi": true,
```
dmijatovic, line 5: - I am not in favor of semi :-)
- I am not in favor of large printWidth, 72 is optimal for me
- I prefer spaces instead of tabs, but mainly the size of 2 instead of 4 is important for me.

ewan-escience: Semicolons prevent hard to detect bugs. If you have a better argument than [this](https://old.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/) for tabs over spaces, I'll reconsider. And with printwidth 80, a lot gets broken down into newlines, which I guess is a matter of taste, but makes it harder to read for me.

I'd actually argue that we use the same settings in the frontend as well.

But this PR is about using Prettier to enforce the existing formatting, not to have an argument on what the formatting should be.

dmijatovic (approved):
Nice work! Thanks

PR #1206, opened 2024-05-23
Increase character limit for mention title

### LINKED ISSUE #1205: Character limit of 500 for mention title too low
The mention with DOI `10.1103/physrevd.108.072003` ([link](https://doi.org/10.1103/physrevd.108.072003)) has a title of 543 characters (see the [Crossref API](https://api.crossref.org/works/10.1103/physrevd.108.072003)), while the database enforces a maximum of 500 characters. We should increase the database limit, perhaps to 1000.

ewan-escience:
## Increase character limit for mention title

* Increase the character limit for mention titles from 500 to 3000 in the database
* Adapt the frontend to this limit
* Set other limits in the frontend in line with the database

How to test:

* Sign in as admin, create a software or project page, add a mention with DOI `10.1103/physrevd.108.072003` and with `10.1103/physrevlett.131.251802`, this should work without an error
* On the mentions overview for admins, when editing this mention, you should see the new limit
* Same when adding a new manual mention
* Also check if the other values in the frontend config are in line with the database

Closes #1205

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/database/009-create-mention-table.sql
+++ b/database/009-create-mention-table.sql
@@ -1,5 +1,5 @@
--- SPDX-FileCopyrightText: 2021 - 2023 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
--- SPDX-FileCopyrightText: 2021 - 2023 Netherlands eScience Center
+-- SPDX-FileCopyrightText: 2021 - 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+-- SPDX-FileCopyrightText: 2021 - 2024 Netherlands eScience Center
 -- SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 -- SPDX-FileCopyrightText: 2022 dv4all
 --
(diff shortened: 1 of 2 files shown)


ewan-escience:
I updated the limit to be 3000 instead of 1000, because the mention with DOI `10.1103/physrevlett.131.251802` went over that limit. Please test with this DOI as well.

dmijatovic:
I have 2 remarks we might address it in this PR or in a new one.

- The authors length of 50000 seem not to work optimally in the modal. The amount of 50.000 chars combined with the form functionality we offer (char count and direct error signaling) breaks at ~ 25.000 chars on my PC. Maybe rsd-admin should have option to use all 50.000 but I would strongly suggest to limit the authors length to 10.000 for other users.
- The author without expected props produces undefined entry. See example 10.1103/physrevd.108.072003. The last author has only `name` prop. We should improve import to use name when "expected" name props are not delivered.

dmijatovic (approved):

ewan-escience:
> The amount of 50.000 chars combined with the form functionality we offer (char count and direct error signaling) breaks at ~ 25.000 chars on my PC.

I think in practice, most entries will have a size well below that. However, I don't know how to solve that for the few exceptional cases. Maybe trade React for something capable. 😉

> The author without expected props produces undefined entry. See example 10.1103/physrevd.108.072003. The last author has only name prop. We should improve import to use name when "expected" name props are not delivered.

Nice find! That's an existing bug, I will make an issue out of it.

PR #1143, opened 2024-03-19
improve software categories

fembau:
- add default categories #992 
- add provenance to link to controlled vocabularies #1141 

### How to test:

see #984

PR #1219, opened 2024-06-04
Add table repr for highlighted categories

### LINKED ISSUE #1217: upgrade tailwindcss package
I'm currently working on the table representation of category hierarchies (#992) and I'm facing an incorrect rendering when using tailwinds [subgrid](https://tailwindcss.com/docs/grid-template-columns#subgrid): Cells inside the subgrid are rendered vertically instead of horizontally. I tried tailwindcss v3.4.3 and it is fixed there.

fembau:
* add table representation for highlighted categories (was separated from #1143 )

Closes #1217


ewan-escience (changes requested):
See the comment from @cmeessen.

dmijatovic:
Yes, definitely something unexpected happening. Further investigation is needed. I'll have a look if I can find the culprit.

cmeessen (changes requested):
For mobile devices, please display the bullet list instead of the table.

fembau:
fixed table rendering and added different representation for small screens

ewan-escience (approved):

cmeessen (approved):
Works as expected, great work :+1:

fembau:
@ewan-escience You replace the basic data structure of category tree by the new class `TreeNode`. I see a lots of
```
const category = item.getValue()
if (category === null) {
    return null
}
```
Why `value` of a tree node is allowed to be `null`. Makes no sense for me :thinking:

ewan-escience:
> Why `value` of a tree node is allowed to be `null`. Makes no sense for me 🤔

Because during initalisation of the tree, I need the value be null sometimes, see e.g . the function `categoryEntriesToRoots` in `frontend/components/category/apiCategories.ts`.

ewan-escience:
I can maybe do without `null` by keeping track of the children of a not-yet-existing node using an extra hashmap. Not having `null` for the value would indeed be a lot more convenient.

PR #1221, opened 2024-06-05
Expire maintainer invites and scheduled tasks for the database

### LINKED ISSUE #260: Maintainer flow improvements
We should consider the following improvements to the maintainer flow:

- invitations should have an expiration date, a maximum validity period (e.g. 1 week) should be enforced
- it should be possible to see invitations (and who accepted them)
- it should be possible to delete/retract unused invitations (implemented in #485)
- expired unused invitations should regularly be cleaned up from the database
- maybe used invitations should be cleaned up as well
- show the error messages from the database on failure instead of the generic message (implemented in #258)

ewan-escience:
## Expire maintainer invites and scheduled tasks for the database

* Maintainer invites now expire 31 days after creation; this is implemented by an automatically generated column in the database
* Show the remaining validity of invites in days on the frontend
* Add documentation on how to schedule repeated tasks in the database

How to test:

* Sign in as admin, create maintainer invites for a software page, project and organisation
* Each invitation should show that it will expire in 31 days
* You can run the following snippet in the database to experiment with different expiration dates for software invites to check if the frontend behaves correctly:
```sql
ALTER TABLE invite_maintainer_for_software DISABLE TRIGGER sanitise_update_invite_maintainer_for_software;
UPDATE invite_maintainer_for_software SET created_at = CURRENT_TIMESTAMP - INTERVAL '31 days';
ALTER TABLE invite_maintainer_for_software ENABLE TRIGGER sanitise_update_invite_maintainer_for_software;
```
* Check the updated docs on this
* Check the new section [here](http://localhost/documentation/rsd-instance/database/#scheduling-repeated-tasks) on running repeated tasks
* Optional: you could run a cron job similar to the ones in the docs by adding the following line to your crontab, make sure to **edit the paths**
```
* * * * * docker-compose --file /home/ubuntu/docker-compose.yml exec -T database psql --dbname=rsd-db --username=rsd --command="SELECT COUNT(*) FROM software;" >> /home/ubuntu/cron.log 2>&1
```

Closes #260

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/database/015-create-maintainer-tables.sql
+++ b/database/015-create-maintainer-tables.sql
@@ -1,5 +1,5 @@
--- SPDX-FileCopyrightText: 2021 - 2022 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
--- SPDX-FileCopyrightText: 2021 - 2022 Netherlands eScience Center
+-- SPDX-FileCopyrightText: 2021 - 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+-- SPDX-FileCopyrightText: 2021 - 2024 Netherlands eScience Center
 -- SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 -- SPDX-FileCopyrightText: 2022 dv4all
 --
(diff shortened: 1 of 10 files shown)


dmijatovic (approved):
Nice work!

cmeessen (approved):
Works as described.

Would it be possible to install cron in the Postgres container so that we can add the scheduled task in the container?

ewan-escience:
> Would it be possible to install cron in the Postgres container so that we can add the scheduled task in the container?

We decided on investigating having a dedicated container for `cron` jobs.

About the 31 day, it is not allowed to use a subquery in a generated column, so we cannot use any env variables to change it. I'm fine with changing it to another value, like 14 days, but I don't think it's possible to make it configurable. You can of yourse change the column manually on your instance.

PR #1243, opened 2024-07-04
Restrict MIME types for image table

### LINKED ISSUE #1241: Allowing arbitrary MIME types for images enables XSS
In `database/003-create-image-table.sql`, in the `image` table, there is a column called `mime_type`. Currently, this column allows for arbitrary data to be stored. If someone stores an HTML file with MIME type `text/html`, this allows arbitrary scripts to be executed when someone 'views' the image. So we should restrict the column to only contain valid image MIME types.

A hotfix can be applied with the following SQL command:

```sql
ALTER TABLE image ADD CONSTRAINT image_valid_mime_type CHECK (mime_type IN ('image/avif', 'image/gif', 'image/jpeg', 'image/png', 'image/svg+xml', 'image/webp'));
```

This was already applied to the RSD and Helmholtz production instances. We will also add this to the codebase. Furthermore, since on each instance the userbase is restricted, we don't expect this to have been exploited in practice.

To see if you need more MIME types, run

```sql
select mime_type, count(*) from image group by mime_type order by mime_type;
```

All image MIME types can be found [here](https://www.iana.org/assignments/media-types/media-types.xhtml#image). @cmeessen, do you need more MIME types other than the ones listed in the hotfix?

ewan-escience:
## Restrict MIME types for image table

* Restrict the allowed MIME types in the `image` table
* Adapt the frontend to show, by default, only allowed files when choosing an image
* Adapt the data generation script
* Remove an unused function from the data generation script

How to test:

* Sign in as admin
* Upload some images in various places
* By default, the upload window will only show accepted files
* If you try to upload a non-accepted file, it should fail with an error

**Note**: you might have the non-existing MIME type `image/jpg` in the database. Before applying this in production, you should run the following SQL query:

```sql
UPDATE IMAGE SET mime_type = 'image/jpeg' WHERE mime_type = 'image/jpg';
```

Closes #1241

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/utils/handleFileUpload.ts
+++ b/frontend/utils/handleFileUpload.ts
@@ -1,7 +1,8 @@
+// SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (dv4all)
-// SPDX-FileCopyrightText: 2023 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2023 dv4all
+// SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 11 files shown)


cmeessen (approved):

frontend/utils/handleFileUpload.ts
```diff
@@ -77,7 +78,7 @@ export function showDialogAndGetFile(): Promise<HandleFileUploadResponse> {
       input.id = id
       input.type = 'file'
       input.name = 'file-input-element'
-      input.accept = 'image/*'
+      input.accept = 'allowedImageMimeTypes'
```
dmijatovic, line 81: This should be without quotes
```
input.accept = allowedImageMimeTypes
```

dmijatovic (approved):
Nice work

PR #1246, opened 2024-07-04
Allow adding community categories to software

### LINKED ISSUE #1212: Categories for communities
A maintainer of a community should be able to create categories specific to that community. Only software that wants to be part of that community should be able to apply these categories to itself. The result should be visible on the public software page.

ewan-escience:
## Allow adding community categories to software

* When joining a community, you are prompted to add the corresponding categories to the software page
* After joining, you can still edit the categories

How to test:

* Sign in as admin
* Join some communities on some software pages
* Play around with adding and removing categories
* Play around with joining and leaving communities
* Check that on the software page, the categories are always correctly shown

**Note**: I'm open for stylistic changes, but changing the workflow would take too much effort/time in this PR, unless something is fundamentally broken. We want to show this to an external partner in a week.

Closes #1212

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests


cmeessen:
Adding and removing the community categories to a software works as described :+1:.

From my point of view, being directly prompted the category selection in a popup the user could be a bit overwhelmed. Also, when adding a community without categories, the user is prompted with a window that there are no categories in this community. I think that is not what the user would expect. In my opinion, and if is not too much work, the popup could be removed because there is still the option to add the categories right after applying to the community.

Other than that, I would add a bit more padding above and below the category lists in the software edit view.

jmaassen:
Works as expected, but the UI is still a bit clunky..

When initially adding a community and categories, the giant popup can be a bit overwelming:

Screenshot_2024-07-05_09-32-19

It was initially unclear to me how to proceed after selecting categories, because the save button is on the smaller popup under the bigger one. Also, when selecting categories, you can see the smaller popup underneath change shape, which is a bit weird.

When going back to the smaller popup, the layout is a bit squashed, as seen here:

Screenshot_2024-07-05_09-32-35

This squashed layout also happens in the regular communities view: 

Screenshot_2024-07-05_09-32-45

What is also a bit inconsistent, is that I can directly delete properties from this 'regular' community view, but when I want to add new ones, I get the large popup again in which I can both add and delete. 

When I've added the categories of a community, these will show up on the software page even if the software has not been accepted into the community yet. So the community badge is not on the page and it's unclear where these catagories refer to?

Screenshot_2024-07-05_09-54-08

When selecting multiple communities, the community tab becomes very busy. More whitespace would help?

Screenshot_2024-07-05_09-52-44

Also, it becomes difficult to see which category belongs to which community on the software page: 

Screenshot_2024-07-05_09-53-10

ewan-escience:
Update: the communities list now doesn't show the categories. Yoy can see and edit the categories when you click the "Edit categories" button. When you join a community that doesn't have any categories, the modal automatically closes and the request to join the community is automatically sent.

We didn't change how the categories are displayed on the software page. If changes are needed, that would be for another branch.

frontend/components/software/edit/communities/CommunityAddCategoriesDialog.tsx
```diff
@@ -0,0 +1,246 @@
...
+        borderBottom: '1px solid',
+        borderColor: 'divider',
+        color: 'primary.main',
+        fontWeight: 500
+      }}>Add categories of {community.name}</DialogTitle>
+      <DialogContent
+        sx={{
+          display: 'flex',
+          flexDirection: 'column',
+          width: ['100%', '37rem'],
```
dmijatovic, line 212: On second thought I think is better to remove width definition ` width: ['100%', '37rem']`
Sorry about this. On large screens the width of modal can be larger than 37rem (it goes up to 40rem). In that case the context is larger than it should be.

dmijatovic (approved):
Nice work! I had few minor suggestions

jmaassen:
Works well. One small comment: 

For editing the categories, a different icons is used compare to all the other "edit" things on the mainainer page: 

Screenshot 2024-07-10 at 15-15-07 Edit software Research Software Directory

Versus: 

Screenshot 2024-07-10 at 15-15-20 Edit software Research Software Directory

Screenshot 2024-07-10 at 15-17-21 Edit software Research Software Directory

Screenshot 2024-07-10 at 15-17-31 Edit software Research Software Directory

It would make more sense to use the same pencil icon for the categories too? This way the UI is more consistent.

dmijatovic:
> Works well. One small comment:
> 
> For editing the categories, a different icons is used compare to all the other "edit" things on the mainainer page:

Yes, I found it an good idea because the user does not edit item (community) but it changes categories of his software. In other cases the users edit the item from the list (contributor in your example). So I think it is correct to use different icon.

cmeessen (approved):

cmeessen:
I do agree with Jason about the edit button, but it can be changed later as well if we get user responses about it.

PR #1252, opened 2024-07-16
Add communities to global search

### LINKED ISSUE #1248: Add communities to the global search
Even when communities are enabled, the global search bar only searches for software, projects and communities. We should also search for communities.

ewan-escience:
## Add communities to global search

* Adapt the `global_search` function in the database to also search for communities
* When there are no search results, the global search shows hotlinks to all enabled modules
* Return the proper index of where the search term is found in the `global_search` database function for certain cases
* Minor impovements to the code

How to test:

* Play around with the global search functionality
* Play around with different modules by changing the `modules` array in `deployment/rsd/data/settings.json` and by mounting that file as a volume in `docker-compose.yml` (do `docker compose up --scale data-generation=0` after changing)

**Note:** The global search bar will also show results from modules that are not enabled. This might be a bit weird while testing, but I think this won't be a problem in practice. It is easy to filter them out, but might be inefficient to do so, since we then have to filter over an ad-hoc computed column.

Closes #1248

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests


jmaassen (approved):
Works as expected.

PR #1279, opened 2024-08-20
rsd admin can edit package manager item

### LINKED ISSUE #1277: Package manager admin messaging
We could improve ui for admins for disabling package manager scrapers.

image

dmijatovic:
# Rsd admin can edit package manager item

Closes #1277

* Rsd admin can edit package manager item. It can: 
    * disable service by adding a comment into designated input field. The input is enabled only for the services that are "supported". **The info about "supported services" is kept locally (in an data object) and need to be manually updated.**
    * change "detected platform" using a drop-down options. The list of supported providers is kept locally (in same data object) and need to be manually updated. 
* When updating package manager entry only platform and the reasons for disabling are updated

How to test:
* `make start` to build app and create test data
* login as rsd_admin, first user should become rsd_admin automatically if you have correct .env settings
* open first software item and navigate to package managers section
* confirm you can edit existing item
* confirm you can provide reason to stop scraping
* confirm that disabling is visible on the PackMan page
* confirm that disabling is visible on the background services page
* confirm you can create new item
* login as regular rsd user
* edit or create software item
* navigate to package manager section
* confirm you cannot edit package manager item
* confirm you can delete and add new items

## Add package manager (not changed)
image

## Admin package manager modal (to disable service or change platform)
image

## Example disabled service on PM page
image

## Example disabled service on services page
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/005-create-relations-for-software.sql
+++ b/database/005-create-relations-for-software.sql
@@ -38,7 +38,9 @@ CREATE TABLE repository_url (
 	scraping_disabled_reason VARCHAR(200)
 );
 
-
+-- NOTE! When changing package_manager_type
+-- please update packageManagerSettings const
+-- in the frontend file apiPackageManager.ts
 CREATE TYPE package_manager_type AS ENUM (
 	'anaconda',
 	'chocolatey',
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -110,7 +110,7 @@ services:
       # dockerfile to use for build
       dockerfile: Dockerfile
     # update version number to correspond to frontend/package.json
-    image: rsd/frontend:2.16.0
+    image: rsd/frontend:2.19.0
     environment:
       # it uses values from .env file
       - POSTGREST_URL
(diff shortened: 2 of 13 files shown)


jmaassen:
Setting the message works well, and the difference between admin and regular user is as expected. 

However, when I set the message as an admin, I don't see any visual feedback on this as a user in the package manager section (only in the 'background services section'). An improvement would be to add some icon or message indicating a error, that links to the background services section?

jmaassen (approved):
Works as expected.

ewan-escience (changes requested):
When the PATCH request is sent, all the data is sent, but that should only be the data in the form.

When the URL is changed, the counts are not reset to `null`. I actually don't think it should be possible to change the URL (instead delete the entry and create a new one).

jmaassen:
Works well. 

Small comment: 

Screenshot 2024-08-23 at 09-38-39 Edit software Research Software Directory

Instead of saying "Why ... is disabled?" it may be better to say "Reason to disable ...."

ewan-escience (changes requested):
Nice work, I have two small issues left:

1. The `id` is still sent in the PATCH request
2. When opening the edit modal, the `Detected platform` field always shows the detected platform, not the platform saved in the database, which is a bit confusing. I would rename it to just `Platform` and show the saved platform (maybe also showing the detected platform separately).

ewan-escience (approved):

PR #1285, opened 2024-08-23
Replace unimported with knip

### LINKED ISSUE #1273: Stop using the unmaintained project unimported
For the frontend, we use the [unimported](https://github.com/smeijer/unimported) project for checking for unused files and imports. However, this project has been archived and is no longer maintained. The author of unimported suggests to use [Knip](https://knip.dev/), which we should investigate. Knip also [compares itself to similar projects](https://knip.dev/explanations/comparison-and-migration), which we could also have a look at.

dmijatovic:
# Replace unimported with knip

Closes #1273

*  Replace "retired" [unimported](https://github.com/smeijer/unimported) with [knip](https://knip.dev/).

How to test:
* This is specific frontend "feature" to check for unimported files, unused types, unused dependencies and some other "quality" checks.  It is not integrated in the CI/CD, rather it is executed manually from time to time. **To test you will need to have Node and NPM installed on your machine.**
* Navigate to frontend folder and run `npm install` to install all dependencies. You might also remove old node_modules folder using `rm -rf node_modules` **before running** `npm install`
* `npm run knip:files` to check for unimported/unused files
* `npm run knip:types` to check for unused exported type definitions
* Both checks should produce 0 in this PR because I removed unused files and types.
* `npm test` to run unit tests locally. All test should pass.
* `npm run build` to build frontend. The build should be successful.
* `make start` to build app and create test data. The app should work without problems.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ewan-escience (approved):
Works well, I have two questions:

1. The command `npm run knip:exports` does show a lot of unused exports, do you plan on fixing those?
2. Why does the `typescript` entry in `frontend/package.json` have a caret (`^`) in its version now? I'd prefer to have a fixed version, or a tilde (`~`) if necessary for patch version updates.

dmijatovic:
@ewan-escience Thanks. I'll fix typescript version, it seem to be changed during install of knip.
Concerning the exports "feature", I tried it but do not see added value at the moment. The exported functions are used in the file itself by some other functions. The reason that these functions are exported is to indicate that these can be used from other location, so the functions are not necessary the private functions.

PR #808, opened 2023-03-28
Mention feedback

ewan-escience:
# Various mention improvements

Changes proposed in this pull request:

* When a search term contains a DOI anywhere (URL or not), it is detected and we will search on that DOI only (#745)
* Bulk import of mentions with a DOI, up to 75 a time (#786)
* Remove double spaces when searching for a mention title (#795)

How to test:

* `docker-compose down --volumes && docker-compose build --parallel && docker-compose up --scale data-generation=0`.
* Login, create a software page, go to mentions and add the following list:
```
https://doi.ieeecomputersociety.org/10.1109/PMBS56514.2022.00010
10.5281/zenodo.3734289doesnotexist
not a DOI
   10.5281/zenodo.3734289   
   10.5281/zenodo.5604609
10.1016/j.sleep.2019.02.008   
10.2139/ssrn.4090562
10.1109/eScience.2016.7870882
10.3389/fmars.2021.667591
10.5281/zenodo.3993242
10.1007/10704282_88
10.5281/zenodo.5788657
10.4233/uuid:4bb38399-9267-428f-b10a-80b86e101f23
10.5281/zenodo.3970975
10.1145/3514197.3551252
10.1016/j.dsr.2020.103427
10.5888/pcd16.180200
10.5281/zenodo.7524315
10.1007/s00198-019-04862-6
10.5281/zenodo.4274773
10.5281/zenodo.46572
10.5281/zenodo.58372
10.5281/zenodo.6010407
10.1088/1748-9326/ac5cf4
10.5281/zenodo.1168441
10.5281/zenodo.4017909
10.5281/zenodo.7274246
10.5281/zenodo.1045194
10.5281/zenodo.7274494
10.1007/s42001-021-00145-5
10.3390/children5100137
10.5281/zenodo.1305971
10.5281/zenodo.4290717
10.5281/zenodo.3686602
10.5281/zenodo.4293357
10.5281/zenodo.4027119
10.5281/zenodo.7010595
10.1103/physrevfluids.7.034501
10.5281/zenodo.4525749
10.5281/zenodo.7277738
10.5281/zenodo.3902723
10.5281/zenodo.7096809
10.5281/zenodo.7111441
10.5194/esd-11-751-2020
10.1016/j.ijmultiphaseflow.2022.104153
10.5281/zenodo.1135095
10.1038/nmeth.3959
10.1128/mSystems.00937-21
10.5281/zenodo.5842611
10.5281/zenodo.7150823
10.5281/zenodo.4242056
[...]

PR #1289, opened 2024-09-03
Fix bulk import mentions from Crossref

### LINKED ISSUE #1270: Bulk importing Crossref mentions fails
When using the bulk import feature to import many Crossref mentions, this fails, because the the env variable `CROSSREF_CONTACT_EMAIL` is not used in the requests, resulting in failed or slow API requests to Crossref. See #808 for examples of mentions to bulk import. The cause seems to be that the function `getCrossrefItemByDoi` in `frontend/utils/getCrossref.ts` does not use the function `addPoliteEmail` in the same file.

ewan-escience:
## Fix bulk import mentions from Crossref

* Do requests to Crossref from the Next.js server instead of the browser, to that the polite pool (using email) can be used
* Use a promise pool (or semaphore) to limit the amount of parallel requests to Crossref.

How to test:

* Sign in, create a software page
* Bulk add Crossref mentions to the page, using the examples from #808; this should work without error

Closes #1270

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [x] Tests

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -130,6 +130,7 @@ services:
       - AZURE_LOGIN_PROMPT
       - AZURE_DISPLAY_NAME
       - AZURE_DESCRIPTION_HTML
+      - CROSSREF_CONTACT_EMAIL
     expose:
       - 3000
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -142,6 +142,7 @@ services:
       - AZURE_LOGIN_PROMPT
       - AZURE_DISPLAY_NAME
       - AZURE_DESCRIPTION_HTML
+      - CROSSREF_CONTACT_EMAIL
     expose:
       - 3000
     depends_on:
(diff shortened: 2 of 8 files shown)


jmaassen (approved):
Works as expected. Some small comments: 

- if there is an empty line in the bulk import list (for example at the bottom), you get a single "not a valid doi" at the top. It may be unclear to users where this comes from. In general, empty or whitespace only lines should be skipped.
- the order of the DOIs changes in the review screen compared to the order in which they where provided. Any reason to do so?
- an import of 50 entries takes about 10 seconds. Long enough for me to start wondering if anything was happening. Would it be possible to show results as they become available? Or a progress bar of some sort?

jmaassen:
With this testing DOI I got the error message "Registration agent not supported": 

10.15941/j.cnki.issn1001-0424.2006.08.009

ewan-escience:
@jmaassen, thanks for the feedback. The module now skips over blank lines. The other points are valid, but are more involved and warrant their own issue I would say.

dmijatovic (approved):
Nice work!

PR #1299, opened 2024-09-13
fix: specify return types in category_path

### LINKED ISSUE #1297: category_path function throws error in a migrated database
The `category_path` function throws an error when trying to view a software page that uses categories. This only occurs when using a migrated database but not in a newly set up database. This is the database log:

```
2024-09-11 10:32:54.201 UTC [294] ERROR:  return type mismatch in function declared to return category
2024-09-11 10:32:54.201 UTC [294] DETAIL:  Final statement returns character varying instead of jsonb at column 5.
2024-09-11 10:32:54.201 UTC [294] CONTEXT:  SQL function "category_path" during startup
    SQL function "category_path_expanded" statement 1
    SQL function "category_paths_by_software_expanded" statement 1
2024-09-11 10:32:54.201 UTC [294] STATEMENT:  WITH pgrst_source AS (SELECT pgrst_call.pgrst_scalar FROM (SELECT $1 AS json_data) pgrst_payload, LATERAL (SELECT "software_id" FROM json_to_record(pgrst_payload.json_data) AS _("software_id" uuid) LIMIT 1) pgrst_body , LATERAL (SELECT "public"."category_paths_by_software_expanded"("software_id" := pgrst_body."software_id") pgrst_scalar) pgrst_call) SELECT null::bigint AS total_result_set, 1 AS page_total, coalesce(json_agg(_postgrest_t.pgrst_scalar)->0, 'null') AS body, nullif(current_setting('response.headers', true), '') AS response_headers, nullif(current_setting('response.status', true), '') AS response_status, '' AS response_inserted FROM (SELECT "category_paths_by_software_expanded".* FROM "pgrst_source" AS "category_paths_by_software_expanded"     ) _postgrest_t
```

`category_path` uses `LIKE` to specify its return type:

```sql
CREATE FUNCTION category_path(category_id UUID)
RETURNS TABLE (LIKE category)
```

The column order in `category` is different in a migrated database, because the `community` column is added to the end of the category table.

cmeessen:
Fixes #1297 

* change column order of `category` table to match table structure of a migrated table
* explicitly specify `category_path` return types

How to test:

* `docker compose build --parallel && docker compose up`
* create categories and assign them to a software
* verify that they are displayed on the software page and that the database container does not generate errors when viewing the page
* restore a database backup of a migrated database (or by checking out v2.18.0, generating data, applying migrations to v2.19.0)
* open database in terminal `docker compose exec -it database psql rsd-db rsd`
* run this in the db terminal
```sql
DROP FUNCTION category_path;

CREATE FUNCTION category_path(category_id UUID)
RETURNS TABLE (
	id UUID,
	parent UUID,
	community UUID,
	short_name VARCHAR,
	name VARCHAR,
	properties JSONB,
	provenance_iri VARCHAR
)
LANGUAGE SQL STABLE AS
$$
	WITH RECURSIVE cat_path AS (
		SELECT *, 1 AS r_index
			FROM category WHERE id = category_id
	UNION ALL
		SELECT category.*, cat_path.r_index+1
			FROM category
			JOIN cat_path
		ON category.id = cat_path.parent
	)
	-- 1. How can we reverse the output rows without injecting a new column (r_index)?
	-- 2. How a table row "type" could be used here Now we have to list all columns of `category` explicitly
	--    I want to have something like `* without 'r_index'` to be independent from modifications of `category`
	-- 3. Maybe this could be improved by using SEARCH keyword.
	SELECT id, parent, community, short_name, name, properties, provenance_iri
	FROM cat_path
	ORDER BY r_index DESC;
$$;
```
* verify that software page with assigned categories shows categories and that no database errors occur upon viewing

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/database/009-create-keyword-and-category.sql
+++ b/database/009-create-keyword-and-category.sql
@@ -2,6 +2,7 @@
 -- SPDX-FileCopyrightText: 2022 - 2024 Netherlands eScience Center
 -- SPDX-FileCopyrightText: 2023 - 2024 Felix Mühlbauer (GFZ) <felix.muehlbauer@gfz-potsdam.de>
 -- SPDX-FileCopyrightText: 2023 - 2024 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
+-- SPDX-FileCopyrightText: 2024 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 --
 -- SPDX-License-Identifier: Apache-2.0
 
@@ -185,7 +186,15 @@ CREATE TRIGGER zzz_check_cycle_categories  -- triggers are executed in alphabeti
 -- helper functions
 
 CREATE FUNCTION category_path(category_id UUID)
-RETURNS TABLE (LIKE category)
+RETURNS TABLE (
+	id UUID,
+	parent UUID,
+	community UUID,
+	short_name VARCHAR,
+	name VARCHAR,
+	properties JSONB,
+	provenance_iri VARCHAR
+)
 LANGUAGE SQL STABLE AS
 $$
 	WITH RECURSIVE cat_path AS (


ewan-escience (commented):
Thanks for the fix. Works well, although I'm not sure about changing the column order in the `category` table. We have many more tables where the order is not correct in this way (and I'm trying to sort the columns in a way in the source code that makes sense for that table).

ewan-escience (approved):

PR #1308, opened 2024-09-30
get plugins configuration suggested changes

dmijatovic:
# Suggested changes

- refactor: get plugins configuration. 
    - we decided to use /plugin/{pluginName}/config route
    - **the plugins are requested on main page load, also when user is not logged in. This approach is preferred for future support of plugins that do not require user login. The plugins that require user login should return empty array (no options).**
- refactor: user menu hook moved inside the component to avoid calls when user is not logged in
- fix: create license files for documentation markdown files
- tests: create svg component test and fix muiSnackbar test

How to test:
*  RSD app should work without any problems. No plugins are defined at this stage.
* modify settings.json to include some imaginary plugins
* rebuild app. 
* Open home page. Confirm that warning is issued in the logs about missing plugins but the RSD app should work without problems
* If you can, test with real plugin that should change user and edit software menus.

PR Checklist:

*   [ ] Update documentation
*   [] Tests


cmeessen (approved):
Thank you very much, everything working as expected

PR #1340, opened 2024-11-07
fix: reserve add slug as used

### LINKED ISSUE #1338: It shouldn't be possible to create a software or project page with slug `add`
If you create a software of project page with slug `add`, you can never view the public page of this item, since `{software,project}/add` shows the form to create a new page.

dmijatovic:
# Used cannot create software or project using "add" slug

Closes #1338 

* Add software and add project validation includes check for "add" slug as this slug is reserved   
* The slug validation feedback to user is improved. From the moment the user start typing/changing slug until the end of the validation we show "loader" in the slug input indicating "pending" validation process. At the same time the save button is disabled.

How to test:
* login to RSD
* create software, try to use "add" as slug. You should get warning that "add" slug is already in use
* create project, tru

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen:
Works well, but I seem to have found one small bug? 

Sometimes when I add software with a duplicate slug, the warning doesn't appear and the loader keeps on turning. No error is shown: 

Screenshot_2024-11-11_11-15-41

As soon as I replaced the name with an unused one, then followed by the used one, the error is shown correctly. 

I can reproduce this with both software and projects.

dmijatovic:
@jmaassen Well spotted! I have fixed the error. Can you test again, please?

ewan-escience (changes requested):
Works mostly well, but I found one small issue.

When editing the slug on an add page, if I delete the slug character by character, fast enough to not lose the spinner, the spinner will run forever when the slug becomes empty.

In the future, it would be nice if the `/add` path would show a page with items you can add instead of a 404 page.

dmijatovic:
@ewan-escience Great! I fixed your remark. Can you validate, please?

ewan-escience (approved):

Issue #1350, opened 2024-11-18
Overview pages per category


ewan-escience:
When clicking on a category on a software or project page, you should be redirected to an overview page where you see all software/projects that have that category.

This can be implemented in two stages:

- [x] Add database functionality that generates the metadata for the overviews.
- [x] Create the overviews on the frontend
     - [x] Create categories filter on organisation project page, done in #1394
     - [x] Create categories filter on the organisation software page, done in #1394
     - [x] Create categories filter on the community software page, done in #1394

PR #1352, opened 2024-11-20
1350 category overview database

ewan-escience:

* Upgrade Postgres from 15.7 to 15.9
* Upgrade PostgREST from v12.2.2 to v12.2.3
* Add database functions that show all categories of software and projects (including categories that are not assigned directly but are a parent of an assigned category)

### How to test

* Sign in as admin, create an organisation, add three levels of categories
* Create a project, publish it, add the organiation, add its lowest level category
* On http://localhost/api/v1/category_for_project?select=*,project(title),category(short_name,name), you should see only one entry
* On http://localhost/api/v1/rpc/projects_of_category?select=*,project(title),category(short_name,name) you should see three entries, as the parents of the category are also shown
* Do the same with a software page, the URLs to check are http://localhost/api/v1/category_for_software?select=*,software(brand_name),category(short_name,name) and http://localhost/api/v1/rpc/software_of_category?select=*,software(brand_name),category(short_name,name) respectively
* Now we will use a populated RSD:
* `docker compose down --volumes && docker compose up --scale data-generation=1`
* http://localhost/api/v1/rpc/projects_of_category and http://localhost/api/v1/rpc/software_of_category should take about half a second to load.
* We should maybe disable [JIT](https://www.postgresql.org/docs/15/jit.html), because those endpoints are faster when JIT is disabled (run `ALTER DATABASE "rsd-db" SET JIT = OFF;` or use the value of `POSTGRES_DB` in your `.env` and then run `docker compose kill -s SIGUSR1 backend`)

Implements the first step of #1350

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/backend-postgrest/Dockerfile
+++ b/backend-postgrest/Dockerfile
@@ -4,4 +4,4 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-FROM postgrest/postgrest:v12.2.2
+FROM postgrest/postgrest:v12.2.3
--- a/database/014-create-keyword-and-category.sql
+++ b/database/014-create-keyword-and-category.sql
@@ -233,6 +233,7 @@ $$;
 -- includes organisation, community and general categories
 -- Note! to filter specific categories of an community or organisation use join with community table
 
+-- if you ever change this table, then software_of_category() needs to be changed as well
 CREATE TABLE category_for_software (
 	software_id UUID REFERENCES software (id),
 	category_id UUID REFERENCES category (id),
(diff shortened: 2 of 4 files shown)


ewan-escience:
I'm a bit afraid that using the `SETOF` return type might cause the same problems as was fixed in #1299, but this was the only way to get resource embedding to work.

dmijatovic (approved):
Nice work! 

I have not testen JIT. Maybe we can do load test using most used endpoints to get better idea about JIT influence?

PR #1361, opened 2025-01-08
1359 linkedin auth

### LINKED ISSUE #1359: Support signing in with LinkedIn
For the KIN project, we should supports signing in with LinkedIn. Documentation can be found [here](https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2).

ewan-escience:

* Add LinkedIn as OpenID provider for signing in
* Add documentation and fix some typos
* Use status code [303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) instead of the default (which happens to be [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302)) when redirecting after signing in, as it semantically and technically better fits the purpose
* Remove unused functions and environment variables from the auth module

### How to test

* Copy the new env variables from `.env.example` to `.env` and fill in the missing values
* Sign in with LinkedIn
* Sign in with other providers to see if this still works

Closes #1359

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests


dmijatovic (approved):
I have tested:
- SURF: it works
- Azure: it works
- ORCID: it works
- LinkedIn: it works
- Local: it works too
- **Helmholtz: it fails**

I think the credentials I am using  for Helmholtz are old/expired and the reason for failure. @cmeessen Can you test Helmholtz login please?

ewan-escience:
Thanks for testing. For me, Helmholtz still works.

PR #1371, opened 2025-01-16
fix: prevent crashes by filtering out null auth providers

### LINKED ISSUE #1367: RSD crashes when one of the auth provider's metadata cannot be fetched
When an auth provider's well-known endpoint cannot be queried normally (e.g. [LinkedIn](https://www.linkedin.com/oauth/.well-known/openid-configuration) gives a `429 Too Many Requests`), the RSD fully crashes with a client side error.

What should happen instead is that the RSD functions as normal, and that, when signing in, the faulty providers are greyed out.

ewan-escience:

* Do an extra null check to prevent crashes when an auth provider is down

### How to test

* Have LinkedIn auth properly setup in your `.env`
* Simulate LinkedIn being down by adding `120.0.0.1 www.linkedin.com` to `/etc/hosts`
* Visit http://localhost/, because of the line in `/etc/hosts`, wait 10 seconds
* Click on Sign in, the LinkedIn option should not be there
* Remove the line from `/etc/hosts`
* Restart the RSD and refresh the page
* LinkedIn should be back as a signing in option
* (Optional): try this out with other auth providers as well

**Note:** In order to get this fix out, I skipped the "greying out" part I suggested in #1367.

Closes #1367

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,7 +12,7 @@
 # If you experience any problems, remove the special characters from the values or place them in quotes (' or ").
 ################ WARNING ################
 
-# .env.local is consumed by frontent (Next)
+# .env.local is consumed by frontend (Next)
 # see https://nextjs.org/docs/basic-features/environment-variables
 
 # .env is consumed by docker-compose.yml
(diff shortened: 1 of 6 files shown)


dmijatovic (approved):
Excellent job!

PR #1374, opened 2025-01-24
fix: optimise GitLab commit history scraper

### LINKED ISSUE #777: Optimise GitLab scraper
The current implementation of the GitLab scraper iterates over all the commits of a repository, every time the scraper runs. For large number of repositories, this is becoming quite inefficient, and generates a lot of unnecessary data transfer.

Here is a proposed mechanism:

By default, the commit history is reverse chronological in order. After every scraping, we store the latest commit id. When the scraper runs the next time, it will only retrieve the data up to this commit and re-calculate the stats based on the previously scraped data. If the scraper does not detect the last stored commit, it will iterate through the whole history and end at the last commit (last in terms of last in the list). The scraper will then re-compute the commit history.

There are other options that we could explore using, e.g. the `since` or `until` attributes (see https://docs.gitlab.com/ee/api/commits.html#list-repository-commits).

cmeessen:
Fixes #777

* optimise gitlab scrapers so that they do not scrape the entire commit history every time they run
* scraper uses `commit_history_scraped_at` to start checking for new commits
* the commit history could become incorrect if
  * there are commits between finishing scraping and writing the timestamp to the database
  * if the entire commit history was rewritten
* added delete buttons to reset scraped metadata (commit history, programming languages, repository statistics)

How to test:

* login and create a new software that uses a gitlab repository url, e.g.
  ```
  https://codebase.helmholtz.cloud/research-software-directory/RSD-as-a-service
  ```
* publish the software entry
* start the commit scrapers
  ```
  docker compose exec scrapers /opt/java/openjdk/bin/java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.git.MainCommits
  ```
* the first time the scraping will take a while until it is finished
* rerun the commit scrapers, this time the scraping should stop 

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/edit/services/ServiceInfoListItem.tsx
+++ b/frontend/components/software/edit/services/ServiceInfoListItem.tsx
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: 2023 - 2024 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2023 - 2025 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2023 - 2025 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 // SPDX-FileCopyrightText: 2025 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 // SPDX-FileCopyrightText: 2025 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
@@ -21,23 +21,27 @@ import IconButton from '@mui/material/IconButton'
 import Tooltip from '@mui/material/Tooltip'
 import ListItemSecondaryAction from '@mui/material/ListItemSecondaryAction'
 import DeleteIcon from '@mui/icons-material/Delete'
-import {deleteServiceDataFromDb} from './apiSoftwareServices'
 import {useSession} from '~/auth'
-import useSoftwareContext from '../useSoftwareContext'
 import useSnackbar from '~/components/snackbar/useSnackbar'
+import useSoftwareContext from '../useSoftwareContext'
+import {deleteServiceDataFromDb} from './apiSoftwareServices'
 
 
-type ServiceInfoListItemProps={
-  readonly title:string
-  readonly scraped_at: string|null
-  readonly last_error: string|null
-  readonly url: string|null
-  readonly platform: CodePlatform|null
-  readonly scraping_disabled_reason: string|null
-  readonly dbprops?: string[]
-}
+type ServiceInfoListItemProps=Readonly<{
+  title:string
+  scraped_at: string|null
+  last_error: string|null
+  url: string|null
+  platform: CodePlatform|null
+  scraping_disabled_reason: string|null
+  dbprops?: string[]
+  onClear?:()=>void
+}>
 
-export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,scraping_disabled_reason,dbprops}:ServiceInfoListItemProps){
+export function ServiceInfoListItem({
+  title,scraped_at,last_error,url,platform,
+  scraping_disabled_reason,dbprops, onClear
+}:ServiceInfoListItemProps){
   let status:'error'|'success'|'not_active'|'scheduled'|'not_supported' = 'not_active'
   const {token} = useSession()
   const {software} = useSoftwareContext()
@@ -97,8 +101,17 @@ export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,sc
   }
 
   async function clearServiceData(dbprops: string[]) {
-    await deleteServiceDataFromDb({dbprops: dbprops, software: software.id, token: token})
-      .then(resp => (resp?.status === 204) ? showSuccessMessage(resp?.message) : showErrorMessage(resp?.message))
+    const resp = await deleteServiceDataFromDb({dbprops, software: software.id, token: token})
+    if (resp?.status === 200) {
+      const formattedProps = dbprops
+        .map((str, index) => (index === 0 ? str.replace(/_/g, ' ').replace(/^./, c => c.toUpperCase()) : str.replace(/_/g, ' ')))
+        .join(', ')
+      showSuccessMessage(`${formattedProps} cleared from database.`)
+      // reload data
+      if (onClear) onClear()
+    }else{
+      showErrorMessage(resp?.message)
+    }
   }
 
   return (
@@ -124,9 +137,9 @@ export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,sc
         secondary={getStatusMsg()}
       />
 
-      { dbprops && (
+      { dbprops ?
         <ListItemSecondaryAction>
-          <Tooltip title={(status == 'not_active') ? 'No repository URL: No data to delete.' : 'Clear service data'}>
+          <Tooltip title={(status === 'success') ? 'Clear service data' : 'No repository URL or no data to delete.'}>
             <span>
               <IconButton
                 onClick={() => {
@@ -134,15 +147,15 @@ export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,sc
                 }}
                 aria-label="delete"
                 size="large"
-                disabled={(status == 'not_active')}
+                disabled={status !== 'success'}
               >
                 <DeleteIcon />
               </IconButton>
             </span>
           </Tooltip>
         </ListItemSecondaryAction>
-      )}
-
+        : null
+      }
     </ListItem>
   )
 }
--- a/frontend/components/software/edit/services/ServiceInfoListItem.tsx
+++ b/frontend/components/software/edit/services/ServiceInfoListItem.tsx
@@ -1,6 +1,9 @@
 // SPDX-FileCopyrightText: 2023 - 2024 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2025 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+// SPDX-FileCopyrightText: 2025 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
+// SPDX-FileCopyrightText: 2025 Paula Stock (GFZ) <paula.stock@gfz.de>
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -14,6 +17,15 @@ import HelpOutlineIcon from '@mui/icons-material/HelpOutline'
 import ScheduleIcon from '@mui/icons-material/Schedule'
 import DoDisturbOnIcon from '@mui/icons-material/DoDisturbOn'
 import {CodePlatform} from '~/types/SoftwareTypes'
+import IconButton from '@mui/material/IconButton'
+import Tooltip from '@mui/material/Tooltip'
+import ListItemSecondaryAction from '@mui/material/ListItemSecondaryAction'
+import DeleteIcon from '@mui/icons-material/Delete'
+import {deleteServiceDataFromDb} from './apiSoftwareServices'
+import {useSession} from '~/auth'
+import useSoftwareContext from '../useSoftwareContext'
+import useSnackbar from '~/components/snackbar/useSnackbar'
+
 
 type ServiceInfoListItemProps={
   readonly title:string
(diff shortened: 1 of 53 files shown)


cmeessen:
I rebased the branch to the current status of the main branch

scrapers/src/main/java/nl/esciencecenter/rsd/scraper/git/BasicRepositoryDataWithHistory.java
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2025 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+// SPDX-FileCopyrightText: 2025 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package nl.esciencecenter.rsd.scraper.git;
+
+import java.util.UUID;
+
+public record BasicRepositoryDataWithHistory(UUID software, String url, String commitHistoryScrapedAt, CommitsPerWeek commitsPerWeek) {
```
ewan-escience, line 10: I would prefer a specialised type for `commitHistoryScrapedAt`, probably [ZonedDateTime](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/ZonedDateTime.html). I see this in other places as well, like `GitlabScraper`. Or do you need strings to make it work with existing code? If so, this can also be changed later.

scrapers/src/main/java/nl/esciencecenter/rsd/scraper/git/CommitsPerWeek.java
```diff
@@ -26,12 +28,23 @@
...
 	static final Gson gson = new GsonBuilder()
 			.enableComplexMapKeySerialization()
 			.registerTypeAdapter(Instant.class, (JsonSerializer<Instant>) (src, typeOfSrc, context) -> new JsonPrimitive(src.getEpochSecond()))
 			.create();
 
+	public SortedMap<Instant, Long> getData() {
+		return new TreeMap<>(data);
+	}
+
+	public void setData(SortedMap<Instant, Long> data) {
```
ewan-escience, line 41: I'm against this method, as the data provided in the parameter might not be consistent with what this class does.

If you really need this (do you?), you should iterate over all entries in the `data` map and add them with the respective `addCommits` method.

scrapers/src/main/java/nl/esciencecenter/rsd/scraper/git/CommitsPerWeek.java
```diff
@@ -26,12 +28,23 @@
...
 public class CommitsPerWeek {
 
-	final SortedMap<Instant, Long> data = new TreeMap<>();
+	private final SortedMap<Instant, Long> data = new TreeMap<>();
 	static final Gson gson = new GsonBuilder()
 			.enableComplexMapKeySerialization()
 			.registerTypeAdapter(Instant.class, (JsonSerializer<Instant>) (src, typeOfSrc, context) -> new JsonPrimitive(src.getEpochSecond()))
 			.create();
 
+	public SortedMap<Instant, Long> getData() {
```
ewan-escience, line 37: Do you really need this other than for testing? If not, it should be package private (i.e. no visibility modifier), but I'm open for arguments for making this public. :)

ewan-escience (changes requested):
Thanks for the pull request! I have some comments:

The delete button for the commit history doesn't clear the `commit_history_scraped_at` field, leading to an incomplete commit history on the next harvest.

Furthermore, I have some comments on the code.

I didn't check the frontend code.

frontend/components/software/edit/services/PackageManagerServices.tsx
```diff
@@ -3,6 +3,7 @@
 // SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
 // SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: EUPL-1.2
```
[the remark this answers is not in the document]
cmeessen: Sorry, that was a mistake. I implemented the code first in our fork and then migrated it to this repository.

frontend/components/software/edit/services/config.ts
```diff
@@ -12,21 +15,25 @@ type ServiceListProps={
...
-  props:{scraped_at:'languages_scraped_at',last_error:'languages_last_error',url:'url'}
+  props:{scraped_at:'languages_scraped_at',last_error:'languages_last_error',url:'url'},
+  dbprops: ['languages']
 },{
   name: 'Repository statistics',
   desc: 'Information is extracted from the repository api (github/gitlab)',
-  props:{scraped_at:'basic_data_scraped_at',last_error:'basic_data_last_error',url:'url'}
+  props:{scraped_at:'basic_data_scraped_at',last_error:'basic_data_last_error',url:'url'},
+  dbprops: ['star_count', 'fork_count', 'open_issue_count', 'contributor_count']
 }]
```
dmijatovic, line 37: I think we should also remove 'last_scraped_at' prop when deleting data. Otherwise looking at the record one might get impression that there is no data (when in fact we deleted it). 

Below is the complete object with added props to dbprops
```typescript
export const repoServiceList:ServiceListProps[]=[{
  name: 'Commit history',
  desc: 'Information is extracted from the repository api (github/gitlab)',
  props:{scraped_at:'commit_history_scraped_at',last_error:'commit_history_last_error',url:'url'},
  dbprops: ['commit_history','commit_history_scraped_at']
},{
  name: 'Programming languages',
  desc: 'Information is extracted from the repository api (github/gitlab)',
  props:{scraped_at:'languages_scraped_at',last_error:'languages_last_error',url:'url'},
  dbprops: ['languages','languages_scraped_at']
},{
  name: 'Repository statistics',
  desc: 'Information is extracted from the repository api (github/gitlab)',
  props:{scraped_at:'basic_data_scraped_at',last_error:'basic_data_last_error',url:'url'},
  dbprops: ['star_count', 'fork_count', 'open_issue_count', 'contributor_count','basic_data_scraped_at']
}]
```

frontend/components/software/edit/services/ServiceInfoListItem.tsx
```diff
@@ -79,6 +96,11 @@ export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,sc
     )
   }
 
+  async function clearServiceData(dbprops: string[]) {
+    await deleteServiceDataFromDb({dbprops: dbprops, software: software.id, token: token})
+      .then(resp => (resp?.status === 204) ? showSuccessMessage(resp?.message) : showErrorMessage(resp?.message))
+  }
+
```
dmijatovic, line 103: I would use one of the approached async/await is the most recent approach. then/catch/finally is original Promises approach. Usually we use one of the approach and not the combination. 

Suggested approach is shown below. Note that I have additional suggestion in which suggested method fits properly.

```typescript
async function clearServiceData(dbprops: string[]) {
    const resp = await deleteServiceDataFromDb({dbprops, software: software.id, token: token})
    if (resp?.status === 200) {
      const formattedProps = dbprops
        .map((str, index) => (index === 0 ? str.replace(/_/g, ' ').replace(/^./, c => c.toUpperCase()) : str.replace(/_/g, ' ')))
        .join(', ')
      showSuccessMessage(`${formattedProps} cleared from database.`)
      // reload data
      if (onClear) onClear()
    }else{
      showErrorMessage(resp?.message)
    }
  }
```

frontend/components/software/edit/services/ServiceInfoListItem.tsx
```diff
@@ -22,10 +34,15 @@ type ServiceInfoListItemProps={
   readonly url: string|null
   readonly platform: CodePlatform|null
   readonly scraping_disabled_reason: string|null
+  readonly dbprops?: string[]
```
dmijatovic, line 37: I suggest to reload the data after delete action. One of the ways is to add method to be called after delete to reload the data.  I propose to add onClear method to type.

Complete ServiceInfoListItemProps type should be

```typescript
type ServiceInfoListItemProps=Readonly<{
  title:string
  scraped_at: string|null
  last_error: string|null
  url: string|null
  platform: CodePlatform|null
  scraping_disabled_reason: string|null
  dbprops?: string[]
  onClear?:()=>void
}>
```

frontend/components/software/edit/services/ServiceInfoListItem.tsx
```diff
@@ -101,6 +123,26 @@ export function ServiceInfoListItem({title,scraped_at,last_error,url,platform,sc
...
+                  clearServiceData(dbprops)
+                }}
+                aria-label="delete"
+                size="large"
+                disabled={(status == 'not_active')}
+              >
+                <DeleteIcon />
+              </IconButton>
+            </span>
+          </Tooltip>
```
dmijatovic, line 142: I think that delete button should be enabled if there is only data to delete. I have suggestion about tooltip messages and disabled button status.

One minor suggestion, use of unnamed function () is not required in logic tests (line 129, 137) and use of exact match '===' (value & type) is prefered.

I provide complete content of suggestion for Tooltip code section
```typescript
<Tooltip title={status === 'success' ? 'Clear service data' : 'No repository URL or no data to delete.'}>
  <span>
    <IconButton
      onClick={() => {
        clearServiceData(dbprops)
      }}
      aria-label="delete"
      size="large"
      disabled={status !== 'success'}
    >
      <DeleteIcon />
    </IconButton>
  </span>
</Tooltip>
```

frontend/components/software/edit/services/apiSoftwareServices.tsx
```diff
@@ -160,3 +163,36 @@ export function useSoftwareServices(){
...
+      }
+    }
+  } catch (e: any) {
+    logger(`deleteServiceDataFromDb: ${e?.message}`, 'error')
+    return {
+      status: 500,
+      message: e?.message
+    }
+  }
+}
```
dmijatovic, line 198: Nice work! 
Few suggestions: we have utility function that extracts database error message, please use it. Currently when patch fails we do not show error message received from api. 
createJsonHeaders adds 'content-type' header too.  

I provide complete function suggestion
```typescript
export async function deleteServiceDataFromDb({dbprops, software, token}:
  {dbprops:string[], software: string, token:string}){
  try {
    const query = `repository_url?software=eq.${software}`
    const url = `${getBaseUrl()}/${query}`
    const resp = await fetch(url, {
      method: 'PATCH',
      headers: {
        ...createJsonHeaders(token)
      },
      body: JSON.stringify(
        dbprops.reduce((acc, dbprop) => ({...acc, [dbprop]: null}), {})
      )
    })
    return extractReturnMessage(resp)
  } catch (e: any) {
    logger(`deleteServiceDataFromDb: ${e?.message}`, 'error')
    return {
      status: 500,
      message: e?.message
    }
  }
}
```

frontend/components/software/edit/services/apiSoftwareServices.tsx
```diff
@@ -160,3 +163,36 @@ export function useSoftwareServices(){
     services
   }
 }
```
dmijatovic, line 165: In order to reload the data after delete, I propose to extend useSoftwareServices hook by adding loadServices method. Then we can use this method after data is delete to reload data. I needed to refactor this hook in order to facilitate new method.

Below it the complete code of useSoftwareServices hook

```typescript
export function useSoftwareServices(){
  const {token} = useSession()
  const {software} = useSoftwareContext()
  const [services,setServices] = useState<SoftwareServices>()
  const [loading, setLoading] = useState(true)

  const loadServices = useCallback((abort:boolean)=>{
    if (token && software.id){
      getSoftwareServices(software.id,token)
        .then(item=>{
          if (abort===false) setServices(item)
        })
        .catch(e=>{
          logger(`useSoftwareServices failed. ${e.message}`,'error')
          if (abort===false) setServices(undefined)
        })
        .finally(()=>{
          if (abort===false) setLoading(false)
        })
    }
  },[token,software.id])

  useEffect(()=>{
    let abort=false

    if (token && software.id){
      setLoading(true)
      loadServices(abort)
    }

    return ()=>{abort=true}
  },[token,software.id,loadServices])


  return {
    loading,
    services,
    loadServices
  }
}

```

frontend/components/software/edit/services/SoftwareRepoServices.tsx
```diff
@@ -30,7 +33,8 @@ export default function SoftwareRepoServices() {
             scraped_at: services ? services[service.props.scraped_at] : null,
             last_error: services ? services[service.props.last_error] : null,
             url: services ? services[service.props.url] : null,
-            platform: services ? services['code_platform'] : null
+            platform: services ? services['code_platform'] : null,
+            dbprops: service.dbprops
           }
           return (
             <ServiceInfoListItem key={service.name} scraping_disabled_reason={null} {...props} />
```
dmijatovic, line 40: In order to facilitate "reload" approach after delete we need to adjust code in this component too. I provide you the content of the complete component because the component is small. Changes are: 
- include loadServices
- use loadServices in ServiceInfoListItem

```typescript
export default function SoftwareRepoServices() {
  const {loading,services,loadServices} = useSoftwareServices()

  if (loading) return <ContentLoader />

  return (
    <>
      {services?.scraping_disabled_reason
        ? <span style={{color: 'red'}}>The harvesters for this repo were disabled by the admins for the following reason: {services?.scraping_disabled_reason}</span>
        : null}
      <List>
        {repoServiceList.map(service=>{
          const props = {
            title: service.name,
            desc: service.desc,
            scraped_at: services ? services[service.props.scraped_at] : null,
            last_error: services ? services[service.props.last_error] : null,
            url: services ? services[service.props.url] : null,
            platform: services ? services['code_platform'] : null,
            dbprops: service.dbprops
          }
          return (
            <ServiceInfoListItem
              key={service.name}
              scraping_disabled_reason={null}
              onClear={()=>loadServices(false)}
              {...props}
            />
          )
        })}
      </List>
    </>
  )
}

```

dmijatovic (changes requested):
@paulastock Nice work for the first PR!

I have some suggestions concerning the frontend code. I would appreciate if you can join one of our standups to discuss these suggestions.

dmijatovic (approved):
Please, double check my frontend suggestions.

cmeessen:
Closing this one due to issues with the history. Moving to #1432.

Issue #1384, opened 2025-01-29
NWO grant IDs


jmaassen:
I was contacted by Tom (with Hans de Jonge in the CC) about the fact that NWO now puts DOIs on their projects. For example: 

​https://doi.org/10.61686/GOLUA29851
https://www.nwo.nl/projecten/golua29851

When I try to add this DOI as a mention, the 'loading icon' shows up and never finishes turning. It doesn't seem to time out either?

ewan-escience:
When you look at the network requests in the browser developer tools, you see we get a response back rather quickly. This means that the response is not what we expect and that the frontend somehow cannot correctly process the response.

jmaassen:
This is our funder ID in crossref by the way:

https://api.crossref.org/funders/100013407

PR #1393, opened 2025-02-07
Prevent mention form from hanging forever

ewan-escience:

* When parsing a Crossref mention, check if the title exists
* Show an error message when something is wrong with a mention
* Remove the unused function `getUrlFromDoiOrg`
* Add a link to some lists of DOIs in the README for developers

### How to test

* Sign in as admin, create a software or project page
* Add as mention `​https://doi.org/10.61686/GOLUA29851`, this should result in an error message concerning its missing title
* Bulk import wasn't changed, this should still result in `Unknown error`
* Check the README

Part of #1384

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/components/mention/FindMentionSection.tsx
+++ b/frontend/components/mention/FindMentionSection.tsx
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
-// SPDX-FileCopyrightText: 2022 - 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-// SPDX-FileCopyrightText: 2022 - 2024 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2022 - 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2022 - 2025 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2024 Dusan Mijatovic (Netherlands eScience Center)
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -16,6 +16,7 @@ import FindMentionInfoPanel from '~/components/mention/FindMentionInfoPanel'
 import useEditMentionReducer from '~/components/mention/useEditMentionReducer'
 import {extractSearchTerm} from '~/components/software/edit/mentions/utils'
 import {getMentionByOpenalexId} from '~/utils/getOpenalex'
+import useSnackbar from '~/components/snackbar/useSnackbar'
 
 type FindProjectMentionProps={
   id:string,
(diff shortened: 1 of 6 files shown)


frontend/pages/api/fe/mention/find_by_title.ts
```diff
@@ -38,7 +38,12 @@ function getUrlForRelationType(relationType: RelationType): string {
  * @returns MentionItem[]
  */
 export async function searchForAvailableMentions({project, searchFor, token, relationType}:
-                                                 { project: string, searchFor: string, token: string, relationType: RelationType}) {
+                                                   {
+                                                     project: string,
+                                                     searchFor: string,
+                                                     token: string,
+                                                     relationType: RelationType
+                                                   }) {
```
dmijatovic, line 46: This is very strange indentation.

frontend/pages/api/fe/mention/find_by_title.ts
```diff
@@ -81,8 +90,13 @@ export async function findPublicationByTitle({project, searchFor, token, relatio
   let crosrefItems: MentionItemProps[] = []
   if (crossref.status === 'fulfilled') {
     crosrefItems = crossref?.value.map((item: CrossrefSelectItem) => {
-      return crossrefItemToMentionItem(item)
+      try {
+        return crossrefItemToMentionItem(item)
+      } catch (e) {
```
dmijatovic, line 95: Please remove e, it is unused
```javascript
try{
  ... 
} catch {
  return null
}

File: frontend/utils/getDOI.ts [hunk 3]
```diff
@@ -74,43 +78,30 @@ export async function getDoiRAList(doiList: string[]) {
...
-        if (item.type.toLowerCase() === 'url') {
-          return item.data.value
-        }
-      }
-    }
-  } catch (e: any) {
-    logger(`getUrlFromDoiOrg: ${e?.message}`, 'error')
-  }
-}
-
```
dmijatovic, line 96: You can also remove code from line 19-37 which is associated with this method

frontend/components/mention/FindMentionSection.tsx
```diff
@@ -59,8 +61,12 @@ export default function FindMentionSection({id,config,findPublicationByTitle}:Fi
         const resp = await getMentionByDoi(searchFor)
         if (resp?.status === 200) {
           return [resp.message as MentionItemProps]
+        } else {
+          if (resp?.message) {
+            showErrorMessage(resp.message)
```
dmijatovic, line 66: Maybe extend/combine error messages. Currently the message is about missing title. 
I think it would be more clear to say: "Failed to import/process mention {DOI}. \n{err.message}"
Something like that.

frontend/pages/api/fe/mention/find_by_title.ts
ewan-escience: What should I put instead? It has to have a name and ESLint also complains when I use `_`.

dmijatovic: What I mean change `catch(e)` to just `catch` because `e` is not used. 
The rest of the code remains unchanged.

PR #1415, opened 2025-03-04
fix: use null instead of empty string value when updating contributor

### LINKED ISSUE #1409: Cannot unset ORCID in admin interface
In the Admin interface of contributors, when an ORCID is set incorrectly it cannot be deleted, only replaced with another ORCID. 

For example, in the entries of Ronald van Haren, the wrong ORCID is used (0000-0003-1250-6968, which is Sven van den Burg). Ronald doesn't seem to have an ORCID himself, and unsetting it gives me: 

Image

dmijatovic:
# Use null instead of empty string for contributor update

Closes #1409

*   When updating contributor data we convert empty string "" to null. This enables removing existing values. 

How to test:
* `make start` to build solution and create test data
* login as rsd admin and navigate to /admin/rsd-contributors page
* add ORCID to an contributor. The value should be saved
* remove ORCID value from the contributor, this should not produce error. You can also validate patch call passing null value.
* You can also remove other values from the contributor like email or affiliation
* Note! You cannot remove first name and last name because these are required values. If you try it you will get error.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/admin/rsd-contributors/apiRsdContributors.ts
+++ b/frontend/components/admin/rsd-contributors/apiRsdContributors.ts
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: 2023 - 2024 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2023 - 2025 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2023 - 2025 Netherlands eScience Center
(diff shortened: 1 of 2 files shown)


ewan-escience (changes requested):
Nice work, but there is still an existing bug:

Start with a contributor without an ORCID. Enter a valid ORCID for them. Then enter an invalid ORCID for them. When trying to save this ORCID, an error appears (which is correct), but the ORCID resets to being empty. Only a refresh will show the correct ORCID, but this has the risk of accidentily deleting the valid ORCID when clicking on that field again.

The same happens when a user starts with a valid ORCID.

The same happens with other fields, like name, as well.

ewan-escience (approved):

PR #1418, opened 2025-03-06
feat: enable removing avatar and enable selection when single image available

### LINKED ISSUE #1410: Cannot select avatar of people with single image in contributor admin interface
When looking at the contributors in the admin interface, I can see the different avatars available for someone: 

Image

When someone is still using the default avatar is some places and has multiple images available, I can select one: 

Image

Once an image is selected, I cannot "unselect" anymore (which is fine): 

Image

However, when someone has a single avatar image, but still uses the default avatar in some places, I cannot select the image. The popup does not appear: 

Image

(so I this case, clicking on the "AvdP" avatar does nothing").

dmijatovic:
# Enable removing avatar on rsd contributor page

Closes #1410

* Enable removing of avatar image of contributor on rsd contributors admin page
* Enable selection of avatar image when only one avatar image available 
* Do not remove changed avatar from RSD. Note! This will produce "dangling" avatar images in RSD database. We will address this by introducing background services in rsd admin section. New issue will be created for this.

How to test:
* `make start` to build app and generate data
* login as rsd admin and navigate to rsd contributors page of admin section
* select contributor with one image. Confirm you can remove avatar.
* find contributor without avatar which has one avatar option. Confirm you can assign avatar.
* find contributor with avatar and multiple options. Confirm you can remove avatar or assign another image

## Example remove avatar from contributor
image

## Example assigning avatar when only one item available
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/admin/rsd-contributors/AvatarOptions.tsx
+++ b/frontend/components/admin/rsd-contributors/AvatarOptions.tsx
@@ -12,8 +12,9 @@ import Badge from '@mui/material/Badge'
 import IconButton from '@mui/material/IconButton'
 
 import {useSession} from '~/auth'
-import {deleteImage, getImageUrl} from '~/utils/editImage'
+import {getImageUrl} from '~/utils/editImage'
 import {getDisplayInitials} from '~/utils/getDisplayName'
+import useSnackbar from '~/components/snackbar/useSnackbar'
 import {RsdContributor} from './useContributors'
 import {patchPerson} from './apiRsdContributors'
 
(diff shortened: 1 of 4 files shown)


ewan-escience (changes requested):
When you remove a contributor image, we try to delete the image itself from the database, to prevent dangling images, which is good.

However, when this succeeds, the counter still shows that that image is available and you can also still use it (the preview of that image is still in memory). However, when you select that image, this fails, since the image is not re-uploaded to the database, only the connection between the contributor and the image is attempted to be restored. So after a successful delete attempt of an image, the image should be re-inserted into the database when trying to revert this removal.

ewan-escience (approved):
Works as advertised.

In response to my previous comment, we decided to not (attempt to) delete images when deselecting them (see also #1419).

A notable missing feature (but this is for another issue), is that you cannot upload a *new* image for a ontributor in the admin view.

PR #1434, opened 2025-03-17
fix: enable delete of parent categories after tree is changed

### LINKED ISSUE #1430: Highlighted category cannot be removed from software if additional child nodes were added
In the software edit view, users cannot remove a category_for_software relation if the admin added more child nodes after the category relation has been added.

### Steps to reproduce

* create categories:
  * 1 - Highlighted
    * 1.1
      * 1.1.1
* reference 1.1.1 in a software entry
* create a new category 1.1.1.1
* go back to software page
* 1.1.1 is still shown on the software page
* open software edit page -> links & metadata
* category 1.1.1 is not shown and association cannot be removed anymore

dmijatovic:
# Enable deletion of parent items when present in data

Closes #1430

* Enable deletion of higher level category item(s) if these are present in the data. The 

How to test:
* create new software entry
* select custom category and confirm it is shown on software page
* now add additional leaf item to category you selected in the admin section of categories
* navigate to software page. Old values are shown. Now go to edit categories page and confirm that delete button is present and you can remove this entry. Note! If you select new item there will be 2 delete buttons because you will need to remove two entries.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/category/CategoriesDialogBody.tsx
+++ b/frontend/components/category/CategoriesDialogBody.tsx
@@ -122,7 +122,7 @@ export default function CategoriesDialogBody({
         filtered = []
         const lowerCaseQuery = searchFor.toLocaleLowerCase()
         for (const root of categories) {
-          const filteredTree = root.subTreeWhereNodesSatisfy(item =>
+          const filteredTree = root.subTreeWhereNodesSatisfyWithChildren(item =>
             item.short_name.toLocaleLowerCase().includes(lowerCaseQuery) ||
             item.name.toLocaleLowerCase().includes(lowerCaseQuery)
           )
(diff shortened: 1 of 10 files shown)


ewan-escience (approved):

cmeessen:
I ran `make start` and edited an existing software. This is what I see under Categories:

image

Is this intentional? The delete buttons are not always on every leaf, sometimes only at the top level.

dmijatovic:
@cmeessen Yes, it is intentional. The delete button is shown near the item that is "saved" in the database. This enables deleting of item which is not at the lowest level but have somehow ended up in the data.

This is "quick" solution I was able to implement within existing approach.

cmeessen:
@dmijatovic I understand. As a user I would assume that the child elements of the selected category are not shown in this list. This is because I would assume that only categories that are assigned to the software will be shown.

dmijatovic:
@cmeessen Ewan contributed improvements to categories approach. Can you check if this approach is more clear?

PR #1438, opened 2025-03-25
feat: add dedicated login page

### LINKED ISSUE #1417: Dedicated login page
We have a use case where a stakeholder would like to provide a link to helmholtz.software that will guide the user directly to a login page. This is desired because the effort for the user to login should be minimised. The current solution would require a lengthy link that contains all the necessary oauth parameters.

I would like to request a dedicated login page that is available via `http://<url>/login`. If there is only one login provider defined, that page would automatically forward the user to the corresponding IDP. If there are multiple providers defined, the page should show the same content as is seen in the login modal on the landing page:

Image

paulastock:
Fixes #1417 

* create a dedicated login page at `pages/login`
* use LoginHandler to handle login action (dialog or redirect) for LoginButton and login page

How to test:

* set `RSD_AUTH_PROVIDERS` in env to only one provider / multiple providers respectively
* running the application, check: 
  * clicking on login button on landing page
  * opening `/login`
* if only one provider is set, this should result in a redirect to the provider URL
* if multiple providers are available, a dialog modal with all provider options should show up

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/login/LoginButton.tsx
+++ b/frontend/components/login/LoginButton.tsx
@@ -1,32 +1,24 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
-// SPDX-FileCopyrightText: 2022 - 2023 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 // SPDX-FileCopyrightText: 2022 - 2024 Netherlands eScience Center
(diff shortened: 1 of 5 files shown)


cmeessen (changes requested):
After logging in the user menu circle is missing. Instead, it shows "Sign in" although I am already signed in and can add software etc.

image

dmijatovic:
@paulastock, @cmeessen I had a quick look at the code. I would like to implement this slightly differently. 
- I would like redirect on login page to be server side when one provider is present. 
- If there are no providers I think we might use 404 page

I am looking now how to implement suggested approach. Can I push new code to this branch on should I create new PR with suggested approach?

PR #1443, opened 2025-03-28
Fix URL to code of conduct

jdswinbank:
I am assuming that a simple typo fix doesn't need test cases, doc updates, etc... hope that's ok!


dmijatovic (approved):
Nice work! Thanks

Issue #854, opened 2023-04-24
Server side render organisation pages


ewan-escience:
Currently, all pages of organisations are client side rendered only (e.g. https://research-software-directory.org/organisations/netherlands-escience-center?page=software and https://research-software-directory.org/organisations/netherlands-escience-center?page=units). We should make this server side rendered, for better SEO and user experience.

PR #1465, opened 2025-04-17
Server side render organisation logo, releases and units

ewan-escience:

* Server side render organisation logo, releases and units
* Adapted some tests

### How to test

* Visit an organisation, disable JavaScript (dev tools, settings, disable JS), the org logo should be shown
* Enable JS, go to the releases, disable JS, all releases should be visible and you can click through the different years
* Changing the year parameter in the URL to something nonsensical will use the highest year as default
* Enable JS, sign in as admin, add some units to the organisation, disable JS, the units should be visible and clickable

Part of #854

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [x] Tests


dmijatovic (changes requested):
I would also prefer additional context for the data.

ewan-escience:
> I would also prefer additional context for new data we are going to obtain server side.

I'm not a fan of contexts. Direct props are easier to reason about and are easier to mock. Moreover, these layers aren't that deep and the data is used only in one place.

dmijatovic:
I am OK with the current approach but I propose to complete other sections before merging into a main. With this approach we will have the complete organisation page working without JS in one release.

jmaassen:
So far it works as advertised. Interesting to see that without Javascript, the software and project tiles do not work for an organization, while they do work for the software / project overview...

I do get an error on the organization overview page: 

```
Content-Security-Policy: The page’s settings blocked a JavaScript eval (script-src) from being executed because it violates the following directive: “script-src 'nonce-533e5623-6bea-4937-bb6c-de431787c1a1' 'strict-dynamic' 'unsafe-inline' https:” (Missing 'unsafe-eval')

inline' https:” (Missing 'unsafe-eval') util.js:1:15666
```

jmaassen (approved):

ewan-escience:
Those errors seem to be unrelated to this PR.

PR #1472, opened 2025-04-28
Include profile entry into contributor/team member search and use account id in the profile url

### LINKED ISSUE #1467: Search for user in user_profile
With introduction of new user profile we should extend search functionality of contributors and team members to include user_profile entries.

In addition, we need to disconnect public profile flag from linking of ORCID account. In this the authentication methods and account linking could go to new section.

dmijatovic:
# Include user_profile in persons search and add support for public user profile page for users without ORCID (via account id)

Closes #1467

* Add account id to contributor and team member tables. Account id is used to match software/project entries for the public user profile page. 
* The user can enable public profile without linking it's ORCID account. The linking to contributor/team member entry is done by account id. When user links ORCID account to it's public profile both account id and ORCID are used to find all contributor and team member entries in the software and projects.
* Public user profile page supports both account id and ORCID for linking user profile to contributor and team member entries in the software and project. By default (internally) the account id is used to generate public user profile page url. However the ORCID url is still supported in case users saved their public profile url (or passed to someone).

How to test:
* `make start` to build and generate test data. Note that generated test data uses same ORCID on the random contributor and team member entries. 
* Login to

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen (approved):

ewan-escience (changes requested):
Nice work.

When your profile is not public yet, the help text says "how to enable?", which doesn't make sense, as you just have to click the button. Furthermore, the documentation it links to is outdated, as it says you need to link your ORCID to be able to enable your public profile.

dmijatovic:
@cmeessen Can you confirm this implementation does not impact Helmholtz implementation? It should not but it is wise to double check :-)

cmeessen:
It works as described, but I did not test the Linkedin login.

What I find confusing is that when a user enables the public profile and is linked in several software entries, this software is only visible in the public profile but not listed in the user settings and vice versa. Furthermore, both views look basically identical which makes it hard to identify what they are currently looking at:

**Public profile**

Public profile

**User settings**

User settings

dmijatovic:
@cmeessen In this PR we "upgraded" user profile UI to be same as public profile. However the content is indeed different, the user profile pages shows software,projects,organisation and communities where the user is maintainer of. The public profile page has different purpose and shows the software and projects where user is mentioned. We will work on extending the content of public profile in the future. I hope this makes current approach more clear.

PR #1473, opened 2025-05-01
1462: RSD admin member invite/registration 

### LINKED ISSUE #1462: Global "member" invite
For the spin-of [KIN project](https://github.com/research-software-directory/KIN-RPD/issues/11) we need the following functionality: Allow login with LinkedIn or ORCID only to invited members.

For this purpose I propose following approach to be implemented in the core RSD

## Global members invite
- This invite is created by RSD admin (in admin section). It expires after X days, where X is defined by rsd admin.
- It allows unlimited number of persons to "join" RSD by using same link and their preferred login methods (all login methods available).
- After user logs in his/her profile is created (already implemented)
- RSD admin can "revoke/delete" invite link at any time from the interface

dmijatovic:
# Member invite / registration

Closes #1462

* **Removes the ORCID whitelist and the env variable RSD_AUTH_USER_MAIL_WHITELIST**
* **env variable values of `RSD_AUTH_PROVIDERS` are changed to support invites for registration**. The provider value contains two parts separated by ":". Example: `ORCID:INVITE_ONLY`. The second value is "accessType" and can be `EVERYONE`, `INVITE_ONLY`, `DISABLED`. `INVITE_ONLY` accessType means that RSD user need to have valid invite link in order to login/register to RSD. After first login the account is added to a list and the user can login next time without invite link.
* RSD invites section is added to RSD admin pages (see image 1). This section enables rsd administrators to create invite/magic link that will register this user and enable his/her access.
* RSD registration page added. This is landing page for the user with the invite link (see image 2). The page list the providers the user can use to register. This list is limited to RSD provider which are defined to support registration/invites.
* Login page and login window updated to use "label" to mark providers with access type "INVITE ONLY". See image 3

How to test:
* update your `.env` file. You can use these values: `RSD_AUTH_PROVIDERS=SURFCONEXT:EVERYONE;AZURE:EVERYONE;LINKEDIN:INVITE_ONLY;ORCID:INVITE_ONLY;LOCAL:EVERYONE`
* `make start` to build and generate data. Note that test data is not required so you can just build and start with docker compose
* login as rsd admin and navigate to RSD invites section.
* Create few different types of invites with different expiration date: 
   *  unlimited number of users
   * 1 user (personal invite)
   * 10 users (group invite)
* Copy this links and use them. You should land on the invite login page that lists all providers that support `INVITE_ONLY` type.
* Test if 1 user invite is allowed to be used only once
* Test if 10 users invite counts down. Not sure if you can test with 10 different users. Change SURFconext to be invite only and use [test users from the user list](https://idp.diy.surfconext.nl/showusers.php)
* Test unlimited number of users with at least 2 users
* Also test provider that supports `EVERYONE` to login. You can also change access type in .env file and test all providers.
* The provider with "accessType" = "DISABLED" should not be listed/shown.

## Example RSD member invite
Expired or used invites have email and copy link buttons disabled.

image

## Example member invite/registration page
We list `INVITE ONLY` providers and Helmholtz (if active) at the registration page because these providers support registration via invite

image

## Updated login screens

We show label for the providers that are `INVITE ONLY` in order to signal that these work only for invited users.
image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


jmaassen:
Works well so far. I do see one issue:

I've got two browser open (firefox and chromium). Admin is logged in on FF, creates invite for 3. Users use invite on chromium and log in via local account. This works, uses gets access and logs out again, next user with different username logs in, etc. Admin sees counter going down, and link expires as expected. 

Once the link is expired however, any __new__ local user trying to log in via chromium get this error message: 

login1

I even get this if I try to log in directly, so not via the invite? 

If I try to log in as a local user on FF I get this: 

login2

which is the expected error in this case?

ewan-escience:
@jmaassen my latest push was an update to delete the account invite cookie after any sign in attempt, did you use that? With that update, the second screenshot is the exptected error message if you don't use an invite after the previous sign in attempt.

ewan-escience (approved):

dmijatovic:
@jmaassen, @cmeessen Can you approve this PR too?

.env.example
```diff
@@ -72,24 +72,21 @@ RSD_ENVIRONMENT=prod
...
 # consumed by services: authentication, frontend (api/fe)
 # provide a list of supported OpenID auth providers
-# the values should be separated by semicolon (;)
-# Allowed values are: SURFCONEXT, HELMHOLTZID, ORCID, AZURE, LINKEDIN or LOCAL
-# if env value is not provided default provider is set to be SURFCONEXT
-# if you add the value "LOCAL", then local accounts are enabled, USE THIS FOR TESTING PURPOSES ONLY
-RSD_AUTH_PROVIDERS=SURFCONEXT;ORCID;AZURE;LINKEDIN;LOCAL
+# the entries should be separated by semicolon (;)
+# Each value should consist of a key and a value (values are CASE SENSITIVE), separated by a colon (:)
+# Allowed keys are: SURFCONEXT, HELMHOLTZ, ORCID, AZURE, LINKEDIN or LOCAL
```
cmeessen, line 77: Is the change from `HELMHOLTZID` to `HELMHOLTZ` on purpose?

dmijatovic: @cmeessen We were unsure which value is "valid". I did include both in the switch statement so both are supported. However I can remove the secondone. If I understand correctly `HELMHOLTZID` is the value you use? Let me know if support for both is OK or you want to have only one value

ewan-escience: It is on purpose, as it needs to be parsable to an `OpenidProvider` enum value, which never had the `id` part in it. @dmijatovic, only `HELMHOLTZ` (case insensitive) is valid now.

frontend/components/login/LoginProviders.tsx
```diff
@@ -7,19 +7,55 @@ import Link from 'next/link'
...
+    )
+  }
+  if (provider?.html){
+    return (
+      <span
+        dangerouslySetInnerHTML={{__html: provider.html}}
+      />
+    )
+  }
+  if (host.email && provider?.accessType==='INVITE_ONLY'){
```
cmeessen, line 40:
```suggestion
  if (host.email && provider?.accessType==='INVITE ONLY'){
```

I would prefer to replace `_` with space as `INVITE_ONLY` appears somewhat like debug output.

.env.example
ewan-escience: We haven't been consistent at all with breaking changes, but go ahead. Also, feel free to commit to this PR. Can you indicate what this fix consists of?

frontend/components/login/LoginProviders.tsx
```diff
@@ -34,17 +70,29 @@ export default function LoginProviders({providers,login_info_url}:LoginProviders
...
-                <ListItemText
-                  primary={provider.name}
-                  secondary={provider?.html ?
-                    <span
-                      dangerouslySetInnerHTML={{__html: provider.html}}
+                {
+                  provider?.accessType==='INVITE_ONLY' ?
+                    <Chip
+                      size='small'
+                      label={provider?.accessType}
```
cmeessen, line 81: This is the right line, as written before I would prefer to replace `_` with spaces, for example like this:

```suggestion
                      label={provider?.accessType?.replace('_', ' ')}
```

.env.example
cmeessen: @ewan-escience when trying to login using an invite, HelmholtzId responds with a "You are not allowed to login" because we are still calling `idUserIsAllowed()`. Right now the user must both have an invite and be in the `HELMHOLTZID_ALLOW_LIST` variable.

dmijatovic: @ewan-escience If this is internal value, can we use `HELMHOLTZID` then instead, concerning the fact that Helmholtz used that value in the past (if I understand the conversation correctly)?

frontend/pages/api/fe/auth/index.ts
```diff
@@ -34,39 +38,102 @@ export type ApiError = {
...
+      if (surf){
+        return {
+          ...surf,
+          accessType
+        }
+      }
+      return null
+    }
+    case 'helmholtz':
+    case 'helmholtzid':{
```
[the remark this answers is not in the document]
dmijatovic: Agree, but let's first decide which value to use. As I see we used `HELMHOLTZID` in the past and all .env variable defining Helmholtz provider use `HELMHOLTZID_` prefix. So, if there is no "historical" impact or strong preference which value to use I would prefer to use `HELMHOLTZID` as we did in the past

frontend/pages/api/fe/auth/index.ts
```diff
@@ -34,39 +38,102 @@ export type ApiError = {
...
   }
 }
-
-async function getProvidersInfo() {
+/**
+ * ServerSide get list of login providers
+ * @param filter
+ * @returns
+ */
+export async function ssrProvidersInfo(filter:ProviderFilter='ENABLED') {
```
cmeessen, line 126: In our case, if `HELMHOLTZ:EVERYONE` is set, the login should still not be allowed for users outside of Helmholtz. However, if `:EVERYONE` is defined, Helmholtz will not appear in the list of login providers when following an invite link. Could this be changed?

frontend/pages/api/fe/auth/index.ts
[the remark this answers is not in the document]
ewan-escience: The `.env` needs to be changed anyways, so let's keep `HELMHOLTZ`. Parsing becomes weird otherwise.

frontend/pages/api/fe/auth/index.ts
dmijatovic: It looks like we might need additional "accessType" beside `EVERYONE` and `INVITE_ONLY`. Let's discuss this on our next weekly and see what should be optimal approach.

ewan-escience: No, the discussion here is going well. ;) @dmijatovic why do you think we need another access type? I think we've covered everything so far. With the HelmholtzID and `HELMHOLTZID_USE_ALLOW_LIST` they've covered their use case. @cmeessen do you think we need another access type?

cmeessen: My understanding was that we were going to remove the environment variables related to the Helmholtz ID allow list so that we can migrate to the new implementation. If this cannot be achieved using the current access types, I'd favour an additional type.

ewan-escience: > My understanding was that we were going to remove the environment variables related to the Helmholtz ID allow list so that we can migrate to the new implementation.

That is not the goal of this PR. The goal is to add a new auth form, namely invites. Furthermore, I removed the ORCID whitelist, because it doesn't work together with the invites. I also got rid of `RSD_AUTH_USER_MAIL_WHITELIST` because it was unused.

We also had the idea to get rid of `HELMHOLTZID_ALLOW_LIST` and put in in the database and integrate it with the frontend, but that was meant for later.

Maybe `HELMHOLTZID_USE_ALLOW_LIST` can be converted into a new access type at some point, but I'm not sure how, as you have *two* different access methods for *one* provider. If you have an idea, can you explain it?

cmeessen (approved):
@dmijatovic thanks for the update, using HelmholtzID works as intended now. I removed unnecessary environment variables, functions and tests.

@ewan-escience `MainTest` is now empty and could be removed.

PR #1476, opened 2025-05-09
fix: clickable mobile menu button

### LINKED ISSUE #1474: Menu item links only work if clicking on text
The links in the menu items in the header

Image

only work if the user clicks on the text itself. If a user instead clicks next to the text, visual feedback suggests forwarding to the page, but nothing happens.

dmijatovic:
# Fix mobile menu items

fix: hide login button when no valid providers

Closes #1474

* fix mobile menu items to be clickable as complete button and mark selected item
* do not show login button when no valid login providers

How to test:
* `make start` to build and generate test data (test data is not required)
* open developer console (F12), select mobile device, the mobile menu will appear. You can click anywhere in the button and the menu should work.
* in your .env file remove all entries from `RSD_AUTH_PROVIDERS` or use invalid values. The login button should not be shown. If you use only one value the login button will forward you directly to login screen of the provider. If there is more than one login provider the modal will appear.

## Example mobile menu

image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/AppHeader/ResponsiveMenu.tsx
+++ b/frontend/components/AppHeader/ResponsiveMenu.tsx
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: 2024 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2024 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2024 - 2025 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2024 - 2025 Netherlands eScience Center
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


ewan-escience (approved):
Works well. Note, the only way to test the sign in button is to supply a wrong value to `RSD_AUTH_PROVIDERS`, as an empty value will use SURFconext by default.

dmijatovic:
> Works well. Note, the only way to test the sign in button is to supply a wrong value to `RSD_AUTH_PROVIDERS`, as an empty value will use SURFconext by default.

Well spotted! I will remove default value so it is possible not to have login option.

ewan-escience:
> I will remove default value so it is possible not to have login option.

That should be part of #1473, where it is disabled on the backend anyways in that case.

cmeessen (approved):

PR #1484, opened 2025-05-21
fix: do not enforce login and consent prompt

### LINKED ISSUE #1483: RSD prevents single-sign-on by enforcing consent prompt
Users in Helmholtz need to consent that RSD wants to access their user data upon every login. The user has to agree every time to this data request:

Consent screen

This behaviour is enforced by `&prompt=login+consent` in `getRedirectUrl()`:

```typescript
export function getRedirectUrl(props: RedirectToProps) {
  let redirectUrl = props.authorization_endpoint +
    '?redirect_uri=' + props.redirect_uri +
    '&client_id=' + props.client_id +
    '&scope=' + props.scope +
    '&response_type=code' +
    '&response_mode=' + props.response_mode
    '&prompt=' + (props.prompt ? props.prompt : 'login+consent')

  if (props?.claims){
    redirectUrl += '&claims=' + encodeURIComponent(JSON.stringify(props.claims))
  }

  return redirectUrl
}
```

However, the `prompt` property cannot be omitted as it is always defined by `getRedirectUrl`.

@dmijatovic Is the default `login+consent` required by any service you are using, or could we make it an optional parameter?

cmeessen:
Fixes #1483 

* do not enforce `prompt=login+consent`

I tested login for Surfconext and HelmholtzID, both still work.

How to test:

* `docker compose build --parallel && docker compose up --scale scrapers=0`
* confirm that login via all IdPs works (especially Linkedin, Azure and Orcid)

PR Checklist:

*   [ ] Update documentation
*   [x] Tests

### DIFF
--- a/frontend/auth/api/authHelpers.test.ts
+++ b/frontend/auth/api/authHelpers.test.ts
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: 2022 dv4all
 // SPDX-FileCopyrightText: 2023 - 2025 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2023 - 2025 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2025 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
+// SPDX-FileCopyrightText: 2025 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 //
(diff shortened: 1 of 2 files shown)


dmijatovic (approved):
Nice work!

I have tested all providers we support except Helmholtz. All work fine. The `prompt` is only used in Azure ISP to show select account window. Other providers do not require it.

Thanks.

ewan-escience (approved):

PR #1485, opened 2025-05-22
1423 access token

### LINKED ISSUE #1423: Access Token for external applications
In theory, the API can be used to push data in an automated way (e.g. CI pipelines) to the RSD. The token that is required for this can only be obtained via interaction with an IDP which requires manual interaction.

To avoid manual interaction and enable full automation, RSD users should be able to create API access tokens that they can use in other applications. These token should be valid for a specified amount of time (e.g. one year), and the user must be able to revoke the access token.

The API access token could be

- universal, meaning write access to all software/projects/organisations where the user is a registered maintainer
- more fine grained, e.g. write access only for specific software

We already discussed two options:

### Access via JWT

- an API endpoint that accepts a "user access token" and returns a short-lived (e.g. 5min) JWT
- pro: full API available

### Dedicated endpoint and service

- a dedicated service with dedicated endpoints that will accept the access token as 
- other metadata formats could be reused (e.g. codemeta, rocrate), but then not all data in the RSD is covered by these schemes

paulastock:
Fixes #1423

* implemented an auth proxy in the authentication module that listens to all requests made to the API and validates access tokens if applicable (for valid requests from the frontend, access token validation is skipped)
   * for requests made to `/api/v2/`
   * this requires a new environment variable `RSD_API_ACCESS_TOKEN_ENABLED` to be set to `true`
   * `RSD_AUTH_URL` now needs to be set to `http://auth:7000/auth`
* added an endpoint for API access tokens in authentication module:
   * POST /auth/accesstoken (create a new API access token)
* added a frontend page in the user settings for managing API access tokens

AccessToken_UI_screenshot_v2

How to test:

* In your `.env`, set `RSD_API_ACCESS_TOKEN_ENABLED=true` and `RSD_AUTH_URL=http://auth:7000/auth` (if not already set, also set `HOST_URL=http://localhost`)
* Start the application and log in as a user
* Go to user settings and select the `API Access Token` tab
* To test access token creation:
   * Enter a token name and expiration date and click on `Create Access Token`
   * Try copying the access token in the then displayed field
   * Try creating a second token for the same user with the same name - this should result in `400 - Failed to create access token. RsdAccessTokenException: Token name should be unique`
   * Try creating a token with an expiration date that is more than 365 days away - this should result in `400 - Failed to create access token. RsdAccessTokenException: Invalid Expiration Date: Access tokens should expire within one year`
* To test access token usage:
   * Try making a request to the RSD API at `/api/v2/` using the Access Token in the Authorization header (`{'Authorization': 'Bearer ACCESS_TOKEN'}`). 
      * For a request that corresponds to the user access rights, it should return a result as normally.
      * For a request of data outside of the user scope, it should not return corresponding data.
* To test access token revocation:
   * Delete a token in the frontend
   * Try making a request to the RSD API at `/api/v2/` using the previously deleted token. It should return `401 - Invalid access token`.
* To test other requests to the RSD API via URL:
   * When using the RSD via the frontend, data should still be available as previously.
   * Requests to the RSD API at `/api/v1/` should still work as before
   * When looking at data from PostgREST via the browser (e.g. `http://localhost/api/v1/software`), public data should still be available

**BREAKING CHANGE: nginx location `/auth/` proxy_pass changed to `http://authentication/auth/`, .env variable `RSD_AUTH_URL` now needs to be set to `http://auth:7000/auth`**

PR Checklist:

*   [x] Update documentation
*   [ ] Tests


authentication/pom.xml
```diff
@@ -198,6 +199,27 @@ SPDX-License-Identifier: Apache-2.0
...
+        <!-- https://mvnrepository.com/artifact/commons-logging/commons-logging -->
+		<dependency>
+			<groupId>commons-logging</groupId>
+			<artifactId>commons-logging</artifactId>
+			<version>1.2</version>
+		</dependency>
+
+        <!-- https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk18on/1.77 -->
+		<dependency>
+			<groupId>org.bouncycastle</groupId>
```
ewan-escience, line 218: Does it not work without this explicit dependency, or, does't Spring Security have the right dependencies itself?

authentication/pom.xml
```diff
@@ -198,6 +199,27 @@ SPDX-License-Identifier: Apache-2.0
...
+        <!-- https://mvnrepository.com/artifact/org.springframework.security/spring-security-crypto -->
+		<dependency>
+			<groupId>org.springframework.security</groupId>
+			<artifactId>spring-security-crypto</artifactId>
+			<version>6.4.5</version>
+		</dependency>
+
+        <!-- https://mvnrepository.com/artifact/commons-logging/commons-logging -->
+		<dependency>
+			<groupId>commons-logging</groupId>
```
ewan-escience, line 211: We already have logback-classic in this module, why was this added?

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Main.java
```diff
@@ -226,7 +237,91 @@ public static void main(String[] args) {
 			});
 		}
 
-		app.get("/refresh", ctx -> {
+		if (Config.isApiAccessTokenEnabled()) {
+			//endpoint to get all access tokes for a user (without secret)
+			app.get("/auth/accesstoken", ctx -> {
```
ewan-escience, line 242: With the RPCs in the database, this is not needed, same for the delete endpoint.

ewan-escience (changes requested):
Without having tested this, I think using the auth module as a proxy for all requests to the PostgREST container is an independent concept from what this PR is trying to achieve and deserves its own issue/discussion/PR (although @jmaassen might think about this differently). Furthermore, I had some initial high level comments.

database/090-access-token.sql
dmijatovic, line 1: Concerning the file name, I would advice to continue with the count: 027-access-token.sql. 
The rule we "try" to follow is the filenames for the tables using continual count and start with 0, the RPC's (functions) start with 1 and, if possible, same second/third digit (example 007-project-tables, 107-project-rpc's. The main reason for file naming order is that some tables and functions depend on tables and functions specified in other file(s).

ewan-escience: I did that on purpose, to make sure there is a gap between the tables and the access token. With later possibly having more fine grained tokens, the token table might have to refer to other tables, which then have to defined before.

nginx/nginx.conf
```diff
@@ -52,7 +52,7 @@ server {
 		add_header Content-Location  /api/$upstream_http_content_location;
 		proxy_set_header  Connection "";
 		proxy_http_version 1.1;
-		proxy_pass http://backend/;
+		proxy_pass http://authentication/api/v1/;
```
dmijatovic, line 55: This seem to be blocking some api calls (remote_rsd, categories, account and maybe more). 

In addition, I am not in favor of proxying all api calls of api/v1 route to auth module. I had different approach in mind. The API processes will request valid JWT from separate auth endpoint passing the access token and the rest of the process remain unchanged.

frontend/components/user/settings/accesstokens/CreateRsdAccessToken.tsx
```diff
@@ -0,0 +1,135 @@
...
+import copyToClipboard from '~/utils/copyToClipboard'
+import {NewAccessToken} from './apiAccessTokens'
+import EditSectionTitle from '~/components/layout/EditSectionTitle'
+import {ContentCopy} from '@mui/icons-material'
+import useSnackbar from '~/components/snackbar/useSnackbar'
+
+
+type CreateRsdAccessTokenProps=Readonly<{
+  createAccessToken:(accesstoken:NewAccessToken)=>string
+}>
```
dmijatovic, line 24: This should be Promise based type (see below). 

I think it would be better to use api call directly in this component without passing it as property (see my comment at onCreateAccessToken

```javascript
type CreateRsdAccessTokenProps=Readonly<{
  createAccessToken:(accesstoken:NewAccessToken)=>Promise<string>
}>
```

authentication/pom.xml
paulastock: Spring Security doesn't have this dependency included by default and it doesn't work without explicitly including it.

authentication/pom.xml
paulastock: Similar to bouncycastle, this dependency was necessary for it to work but wasn't included by default in Spring Security.

ewan-escience (changes requested):
Nice work, in general it seems to work well. I've left a few questions and remarks, along with the remarks below. I haven't looked in-depth at the auth code, I will keep that for the next review.

When I used a deleted access token (both with a PATCH and a GET), I got back an HTML page stating "Unfortunately, something went wrong during the login process.".

When you delete a just created token, its content is still visible in the copy box.

database/090-access-token.sql
```diff
@@ -0,0 +1,94 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+CREATE TABLE user_access_token (
+	id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+	secret VARCHAR NOT NULL,
+	account UUID REFERENCES account (id) NOT NULL,
+	expires_at TIMESTAMPTZ NOT NULL,
+	display_name VARCHAR(100) NOT NULL,
+	created_at TIMESTAMPTZ NOT NULL,
+	UNIQUE(account, display_name)
```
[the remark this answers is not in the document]
paulastock: A unique name seems to be a common requirement for API access tokens (e.g. on GitHub) and it is also a requirement that I would prefer as a user. But you're right about the white spaces - we could trim leading and trailing white spaces on the frontend to avoid this?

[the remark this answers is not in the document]
paulastock: We decided to remove the unique constraint and show the token id instead.

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Main.java
```diff
@@ -230,7 +241,73 @@ public static void main(String[] args) {
...
+				} catch (RsdAccessTokenException e) {
+					ctx.status(400).result(e.getMessage());
+				}
+			});
+
+			app.before("/api/v2/*", ctx -> {
+				String frontendTokenToVerify = ctx.cookie("rsd_token");
+				String authHeader = ctx.header("Authorization");
+				//if request from frontend, skip access token validation
+				if (frontendTokenToVerify != null) {
```
ewan-escience, line 268: What is the purpose of this `if` block? Do you expect the frontend to send requests to `/api/v2`?

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Main.java
```diff
@@ -230,7 +241,73 @@ public static void main(String[] args) {
...
+					String signingSecret = Config.jwtSigningSecret();
+					JwtVerifier verifier = new JwtVerifier(signingSecret);
+					verifier.verify(frontendTokenToVerify);
+					String referer = ctx.header("Referer");
+					if (referer != null && referer.startsWith(Config.hostUrl())) {
+						ctx.attribute("X-API-Authorization-Header", "Bearer " + frontendTokenToVerify);
+					} else {
+						throw new HttpResponseException(401, "Untrusted referer");
```
ewan-escience, line 276: Exceptions like this might be the cause of unwanted redirects. I think it's better to just handle the response with the `ctx` object and return, or throw a self-defined exception and make sure a handler exists for that exception.

ewan-escience (changes requested):
Nice improvements, I have only a few comments remaining. (And I will probably refactor some things afterwards, but that's for later.)

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Main.java
```diff
@@ -386,4 +473,47 @@ static void setRedirectFromCookie(Context ctx) {
...
+		HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+				.uri(URI.create(fullUrl));
+
+		if (ctx.attribute("X-API-Authorization-Header") != null) {
+			requestBuilder.header("Authorization", ctx.attribute("X-API-Authorization-Header"));
+		}
+
+		ctx.headerMap().forEach((k, v) -> {
+			if (!k.equalsIgnoreCase("host") && !k.equalsIgnoreCase("content-length") && !k.equalsIgnoreCase("authorization") && v != null) {
+				requestBuilder.header(k, v);
```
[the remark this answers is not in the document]
paulastock: Good point, I added a function for checking for forbidden header names. This might not cover every forbidden header yet, but can be expanded later.

authentication/src/main/java/nl/esciencecenter/rsd/authentication/Main.java
paulastock: It was a leftover from before we switched to `/api/v2`, I have now removed it since we don't have that use case anymore.

ewan-escience (approved):

jmaassen (approved):
Works as expected. 

I had one edge case where I couldn't add software (as admin) before accepting the TOS & PP. But that is an artifact of the test setup I think.

cmeessen:
For me, tests fail locally at a different spot:

```
$ git rev-parse HEAD
b1fe491cae671a3f980c147d5d21a35aac43decc
$  docker compose up
WARN[0000] /home/USER/git/RSD-as-a-service/e2e/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
[+] Running 1/1
 ✔ Container e2e-tests-1  Created                                                                                                                                   0.0s 
Attaching to tests-1
tests-1  | 
tests-1  | > rsd-e2e-tests@1.0.0 e2e:docker
tests-1  | > playwright test --project=chrome --project=firefox --project=msedge
tests-1  | 

tests-1  | TimeoutError: locator.click: Timeout 30000ms exceeded.
tests-1  | Call log:
tests-1  |   - waiting for getByRole('button', { name: 'Decline' })
tests-1  | 
tests-1  | 
tests-1  |    at ../helpers/globalSetup.ts:50
tests-1  | 
tests-1  |   48 |     await page.getByRole('button', {
tests-1  |   49 |       name: 'Decline'
tests-1  | > 50 |     }).click()
tests-1  |      |        ^
tests-1  |   51 |   }
tests-1  |   52 |   // sign in - we need to do this because we loose cookies
tests-1  |   53 |   await loginLocal({
tests-1  |     at globalSetup (/app/helpers/globalSetup.ts:50:8)
           
tests-1  | To open last HTML report run:
tests-1  | 
tests-1  |   npx playwright show-report
tests-1  | 
tests-1 exited with code 1
```

paulastock:
For me, tests passed locally:

```
e2e % npm run e2e:chrome

> rsd-e2e-tests@1.0.0 e2e:chrome
> playwright test --project=chrome --headed


Running 22 tests using 6 workers
  Slow test file: [chrome] › software.spec.ts (37.2s)
  Slow test file: [chrome] › project.spec.ts (33.5s)
  Consider splitting slow test files to speed up parallel execution
  22 passed (42.2s)
```

ewan-escience:
So if you cannot find a fault in the code, go ahead and merge it. Just beware that all future PRs might have the same problem until the underlying problem is fixed, but I'm ok with that. Or maybe indeed the test is flaky and we wont necessarily see failing tests in other PRs. And could you indeed create an issue about this test?

paulastock:
After updating the `.env.e2e`, e2e test is now passing successfully.

PR #1496, opened 2025-05-31
Obtain OpenID providers from auth module

### LINKED ISSUE #1477: Obtain active OpenID providers from auth module
Currently, both the auth and frontend modules parse the env variable `RSD_AUTH_PROVIDERS` to determine which OpenID providers are enabled. This leads to parsing duplication and possibly inconsistensies. Instead, the auth module should do the parsing and expose an endpoint (for the frontend to consume) that lists all enabled providers.

### LINKED ISSUE #1491: Add login+consent prompt to SURFconext Idp when in 'development' or 'docker' mode
We removed default prompt value `login+consent`. This is desired behaviour in production mode but in development mode it is desired to have prompt enabled. When enabled in development we can change the users and login as different user. Without prompt SURFconext will always return logged in user.

Something like code bellow should be added to `surfcontext.ts` and other providers to enable login screen in dev mode

```javascript
// always show login+consent in dev mode
const prompt = process.env.NODE_ENV === 'development' ? 'login+consent' : process.env.SURFCONEXT_LOGIN_PROMPT
```

**Maybe we need to introduce SURFCONEXT_LOGIN_PROMPT env variable for every provider?**

ewan-escience:

* Create a new endpoint `/auth/providers` that lists all active OpenID providers
* Adapt the frontend to use this endpoint and remove unused code
* Remove unused env variables

### To do

Considering that this PR is already quite large, the following will be done later:

* Obtain the coupling providers from the auth module

### How to test

* Sign in with all providers, this should still work
* Visit http://localhost/login, the list should be the same as the sign in modal
* If `RSD_ENVIRONMENT` is set to `dev`, you will always be forced to provide username and password when using SURFconext, otherwise, single sign-on will be used
* Create a user invite as admin, sign out, visit the invite URL, the list should contain all providers configured as `INVITE_ONLY` (and Helmholtz ID when it is `EVERYONE` as well).
* Linking accounts should still work as usual
* This all should also work in frontend dev mode (`npm run dev`, port 3000)
* Check that the removed env variables are indeed unused

### Open questions

* Should the providers be sorted in the frontend in some specific way? They are currently sorted by when they return from querying the well-known URL.

Closes #1477

Closes #1491

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [x] Tests


dmijatovic (approved):
Nice work!

I left one minor suggestion :-)

authentication/src/main/java/nl/esciencecenter/rsd/authentication/RsdProviders.java
```diff
@@ -0,0 +1,344 @@
...
+				if (Config.isDevEnv()) {
+					queryParameterBuilder.addQueryParameter("prompt", "login");
+				}
+
+				yield queryParameterBuilder.toString();
+			}
+			case helmholtz -> new QueryParameterBuilder()
+				.addQueryParameter("response_mode", "query")
+				.addQueryParameter("response_type", "code")
+				.addQueryParameter("scope", "openid profile email eduperson_principal_name")
```
cmeessen, line 312: For my understanding: why are you hardcoding these parameters when we have the env variable `HELMHOLTZID_SCOPES`? The same holds for the other auth providers.

paulastock (approved):

PR #1506, opened 2025-06-03
chore: scale email services to zero by default

### LINKED ISSUE #1502: Scale mail and publisher services to zero by default
When starting the RSD, the mail and published container are also started by default. These are not used in most cases when developing, and the mail container even crashes indefinitely if you don't have your env variables set up. Therefore, these services should have zero instances by default (use [`replicas`](https://docs.docker.com/reference/compose-file/deploy/#replicas)).

ewan-escience:

* Scale the three email services to zero by default

### How to test

* Check if everything still works
* Check with `docker ps` that the email containers are not present

Closes #1502

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -225,55 +225,6 @@ services:
     networks:
       - net
 
-  rabbitmq:
-    image: rabbitmq:4.0
-    hostname: rabbitmq
-    expose:
-      - "5672"
-    networks:
-      - net
-    volumes:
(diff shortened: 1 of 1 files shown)


dmijatovic (approved):
It works as described.

Maybe we could move these at the bottom of the file and make "section" comment like "Disabled by default in development"

PR #1540, opened 2025-06-18
Update of Imperial landing page

dalonsoa:
# Update of Imperial landing page

Fixes https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/30
Fixes https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/24

Partly addressing new Imperial branding. 

## Changes proposed in this pull request

This PR makes some usability improvements of Imperial's landing page and adapts it to the new Imperial branding, so the right colours can be set to the right things via the `settings.json` file. 

## How to test:

Just use a settings file that sets `imperial` as the host name. The colours and logo will be different, but the changes in functionality, indicated in the two issues above, should be reproducible. Attached a short video showcasing the result. 

https://github.com/user-attachments/assets/680baf31-e7c9-49ac-8d8f-18368a573436

## PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/home/imperial/CounterBox.tsx
+++ b/frontend/components/home/imperial/CounterBox.tsx
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2023 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2025 Diego Alonso Alvarez (Imperial College London) <d.alonso-alvarez@imperial.ac.uk>
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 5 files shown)


jmaassen:
I noticed that the keyword list on the landing page also includes keywords of software that isn't published yet? As a result, clicking on the keyword will (possibly) show you an empty software overview page. 

The keyword filter on the software overview page does take the visibility of the software into account. Maybe you can check how that is implemented?

dalonsoa:
@jmaassen I've removed the token from the api call, which was making keywords related to unpublished software available. Now it only shows the keywords of the software which is publicly available, matching the behaviour of the software submissions label. 

Now this should be ready, hopefully.

dalonsoa:
It now works as I expected, specially after @jmaassen comment, so from my point of view is ready to merge. I cannot do it myself.

PR #1543, opened 2025-06-20
Not accepting terms and conditions sends you to homepage

### LINKED ISSUE #1487: Not accepting terms and conditions sends you back
When clicking the cancel button when prompted for accepting the terms and conditons, you are sent back one page in your browser history. This is strange if you just signed in, as it will send you back to the OpenID provider, and that may not always work. Instead, it would be better to send the user back to the last visited page on the RSD or to the homepage.

ewan-escience:

* When clicking the cancel button on the user agreements modal, you are sent to the homepage instead of one page back, as that might send you to the sign in provider (SURFConext, LinkedIn, etc.)

### How to test

* Sign in (if dev mode, sign in again with a new account, so you are not an admin with the second account)
* Click cancel on any page with the user agreements modal, you will be send to the homepage instead of one page back in your history

closes #1487

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests


jmaassen (approved):
Works as expected. 

Currently, every action you will try to do that requires being logged in will bring you back to the terms and conditions pop-up. As an alternative one could also be logged out instead?

PR #1549, opened 2025-06-26
Obtain coupling providers from backend

### LINKED ISSUE #1517: Obtain active OpenID *coupling* providers from auth module
Just as in #1477, the data for the coupling providers should be obtained from the auth module.

### LINKED ISSUE #1545: Get rid of redirect env variables
Env variables like `ORCID_REDIRECT` and `ORCID_REDIRECT_COUPLE` make it seem like these end points are configurable, while they are in fact not. Now that we have `HOST_URL`, which determines the rest, these variables can be removed.

ewan-escience:

* The frontend now obtains coupling providers from the auth module, instead of parsing environment variables itself and duplicating logic
* Remove env variables that actually cannot be configured freely
* Allow coupling to all active OpenID providers
* Change the SURFconext response mode from `form_post`to `query` because we had trouble [again](https://github.com/research-software-directory/RSD-as-a-service/pull/1473) with `SameSite=Lax` cookies
* Some refactorings to reduce code duplication

### How to test

* Add the necessary `http://localhost/auth/couple/<provider>` redirect URLs to your OpenID apps (not having access to these is not part of this PR)
* Sign in using any method
* Go to your settings, the list of coupling providers should be the same as the login providers, providers that you already coupled should be inactive
* Try any combination of creating new accounts and coupling with different providers
* Check if the removed env variables are correct

closes #1517
closes #1545

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests


frontend/components/user/settings/profile/LinkAccounts.tsx
```diff
@@ -8,45 +9,46 @@
...
-            <LinkAccountBtn
-              disabled={linkedIn!==null}
-              href={linkedInAuthLink}
-              label='Link my LinkedIn'
-            />
-            : null
-          }
-        </div>
-      </div>
-    )
```
dmijatovic, line 48: - You can use `useLoginProviders` hooks to obtain providers list.
- Can we filter out local account type. I do not see a point of having local account as an option
- I would not show account linking section if less than 2 providers are listed
- I would not use List and ListItem components to show buttons, just use div or section with grid class 
- I think it is sufficient to list only provider name in the button. This avoids any lingual/grammar "challenges" when concatenating static text and the provider name.

Here is the content of the complete file LinkAccount.tsx

```javascript
import useLoginProviders from '~/auth/api/useLoginProviders'
import {useUserContext} from '~/components/user/context/UserContext'
import LinkAccountBtn from './LinkAccountBtn'
import {findProviderSubInLogin} from './apiLoginForAccount'

export default function LinkAccounts() {
  const {logins} = useUserContext()
  const {providers} = useLoginProviders()

  // console.group('LinkAccounts')
  // console.log('logins...',logins)
  // console.log('providers...',providers)
  // console.groupEnd()

  if (!providers || providers?.length < 2) {
    return null
  }

  return (
    <div>
      <h3>Link your other accounts</h3>
      <div className="grid gap-4 py-6">
        {providers
          // exclude local account from linking (only for test purposes anyway)
          .filter(provider=>provider.openidProvider!=='local')
          // show all other providers (used will be disabled)
          .map(provider => {
            return (
              <LinkAccountBtn
                key = {provider.openidProvider}
                disabled = {findProviderSubInLogin(logins, provider.openidProvider) !== null}
                href = {provider.coupleUrl}
                label = {provider.name}
              />
            )
          })}
      </div>
    </div>
  )
}

```

frontend/components/user/settings/profile/LinkAccounts.tsx
```diff
@@ -8,45 +9,46 @@
...
+      <List>
+
+        {linkProviders.map(provider => {
+          return (
+            <ListItem key = {provider.openidProvider}>
+              <LinkAccountBtn
+                disabled = {findProviderSubInLogin(logins, provider.openidProvider) !== null}
+                href = {provider.coupleUrl}
+                label = {`Link my ${provider.name} account`}
+              />
```
dmijatovic, line 47: With the current approach we can have up to 4 buttons. I would not use primary button in this case, I would like to propose more "neutral" button type.

The content of LinkAccountBtn component

```javascript
import Button from '@mui/material/Button'

type LinkAccountBtnProps=Readonly<{
  label: string
  disabled: boolean
  href?: string|null
  sxStyle?: any
}>

export default function LinkAccountBtn({label,disabled,href,sxStyle}:LinkAccountBtnProps) {
  if (href){
    return (
      <Button
        disabled={disabled}
        href={href}
        variant="outlined"
        sx={sxStyle}
      >
        {/* Link my account */}
        {label}
      </Button>
    )
  }
  return null
}
```

ewan-escience:
> You can use useLoginProviders hooks to obtain providers list.

I'd have to look into that.

> Can we filter out local account type. I do not see a point of having local account as an option

I think it's weird that you can then couple everything except your local account. Especially since you can couple a local account to other accounts but not vice versa.

> I would not show account linking section if less than 2 providers are listed

Good point.

> I would not use List and ListItem components to show buttons, just use div or section with grid class

I don't know, at least this makes the layout look good. Your example uses Tailwind, of which I thought we should minimise.

> I think it is sufficient to list only provider name in the button. This avoids any lingual/grammar "challenges" when concatenating static text and the provider name.

I'll have a look at this.

dmijatovic (changes requested):
Nice work!

I have two remarks. I provided the content of the two components that I would like to change in order to produce layout as in this print screen

image

frontend/components/user/settings/profile/LinkAccounts.tsx
ewan-escience: I think the outlined buttons look like cancel buttons, I prefer the filled ones. @jmaassen @cmeessen @paulastock do you have an opinion?

dmijatovic (approved):

PR #1551, opened 2025-07-02
Reject expired access tokens

ewan-escience:

* The auth module now rejects expired access tokens

### How to test

* Sign in, create an access token
* Send a GET request to `http://localhost/api/v2/software?limit=10` without a token, this should show software
* * Send a GET request to `http://localhost/api/v2/software?limit=10` With header `Authorization: Bearer <token>`, this should show software
* Run `docker compose exec database psql --dbname=rsd-db --username=rsd --single-transaction --command="ALTER TABLE user_access_token DISABLE TRIGGER sanitise_update_user_access_token; UPDATE user_access_token SET expires_at = NOW(); ALTER TABLE user_access_token ENABLE TRIGGER sanitise_update_user_access_token;"` to expire your access token now
* Repeat the last request, you should get an error message

PR Checklist:

* [ ] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests


dmijatovic (approved):
Nice work! It works as described.

dmijatovic:
@ewan-escience One minor related remark. If I look at the UI it says expires in 0 days. 

I think it would be more clear if we do similar approach as in invites, we show [EXPIRED] in the title. Alternatively we can change expires in 0 days into EXPIRED. In addition, I would put #ID in the new line.

This is not directly related to this PR but it is indirectly related and I wonder if it can be fixed quickly in this PR? Otherwise I wil make a separate issue (see printscreen)

image

PR #1562, opened 2025-07-07
Expand Prettier to Java files

ewan-escience:

* Expand the existing Prettier setup from the data-generation module and move it to the root to include all Java files
* Adapt the GitHub action that checks for correct formatting

### How to test

* Install Prettier in the root of the project (`npm install`) for future contributions
* Test if the RSD still works
* Check the first commit of this PR

Note that the config can still be overridden with different options for other directories

Part of #915

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests


ewan-escience:
> Do we need to do something about Sonar messages?

No, those are existing issues.

PR #1583, opened 2025-07-28
Lock accounts

### LINKED ISSUE #1480: Lock accounts
Admins should be able to (temporarily) lock accounts, in case an account shows malicious behaviour. This should prevent the user from having any write access to the database and maybe read access as well from e.g. unpublished pages.

ewan-escience:

* Allow admins to lock accounts as a security measure
* Allows the admins to set a message the locked user can see and a message that only admins can see
* Locked accounts can not use the API because a pre-request hook will throw an exception

### How to test

* Sign in as admin
* Sign in a as a regular user, copy your rsd_token from the cookies for later use
* Sign in as admin, lock the other user
* When signing in as the regular user again, you should see an error message stating that the account is locked
* When trying to use your token you used earlier (header `Authorization: Bearer <token>`), you should get a `403` with an error object

closes #1480

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [ ] Tests

### DIFF
--- a/documentation/docs/03-rsd-instance/03-administration.md
+++ b/documentation/docs/03-rsd-instance/03-administration.md
@@ -70,6 +70,8 @@ We recommend to use short-lived, single-use invites whenever possible, as these 
 
 This section shows all RSD users who logged in to RSD at least once. You can search for users, assign the administrator role (`rsd_admin`) or delete user accounts.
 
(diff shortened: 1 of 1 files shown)


dmijatovic (approved):
Nice work. I would prefer if editing would be in the modal instead of below list entry but let se what others think :-)

jmaassen (approved):
Works as advertised! 

I do agree that the approach is somewhat out of tune with the other pages where data needs to be entered, but this can also be fixed later?

PR #1586, opened 2025-07-31
feat: use lock account modal for lock info

dmijatovic:
# Use Lock account modal

* Show modal when lock button is clicked to change account lock status
* Validation on lock input is set to min 10 chars and max 100 chart. 
* Providing reasons when account is locked is required
* When account is locked the lock icon is red  

How to test:
* login as rsd admin and navigate to rsd users section
* Try to lock and unlock an account

PR Checklist:

*   [ ] Update documentation
*   [x] Tests

### DIFF
--- a/frontend/components/admin/rsd-users/LockUserModal.tsx
+++ b/frontend/components/admin/rsd-users/LockUserModal.tsx
@@ -1,0 +1,159 @@
+// SPDX-FileCopyrightText: 2025 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2025 Netherlands eScience Center
+//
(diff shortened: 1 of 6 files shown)


ewan-escience (changes requested):
> Validation on lock input is set to min 10 chars and max 100 chart.

I don't agree, it is up to the admin to supply a message of any length, as long as it fits into the database, so the minimum of 10 should be removed.

> Providing reasons when account is locked is required

I don't agree either. You might want to lock an account very quickly in an emergency, in which case having to fill in two text fields might lose you valuable seconds. The reasons can always be added later when necessay.

Furthermore, when trying to remove the lock and when the form is not valid, the save button is disabled.

PR #1590, opened 2025-08-04
feat!: extend modules definitions in settings.json

### LINKED ISSUE #1582: Documentation "Host definitions" is outdated
The documentation about [Host definitions](https://research-software-directory.org/documentation/rsd-instance/configurations/#host-definitions) is outdated, e.g. the news module is not mentioned.

From my point of view, the modules should be explained in a dedicated sub-chapter, as this is becoming more complex. E.g. what does it mean if a module is activated/disabled and where are the modules shown.

dmijatovic:
# Extend modules definition in settings.json

Closes #1582

Question
The news module is not included in global search while all other modules are. I assume this is because the content of the news item is different from other modules, however the question is do we want to include news in global search and what would be optimal approach?

* Remove modules array from host definition
* Introduce new `modules` definition that enables customization of active modules and main menu items label
* Update documentation about changes in settings.json
* Active module values are cached on the frontend in the production mode.

How to test:
* `make start` to build and generate test data.
* Default modules you should see are: software, projects, organisations, communities and news
* Change values in settings.json. 
   * Disable software module. Restart docker (docker compose down && docker compose up). Confirm that software and communities modules are not shown and not accessible (you should see 404 page when visiting software and communities routes). Login to rsd and confirm that menu option add software is not visible.
   * Disable projects module. Restart docker. Confirm that projects overview is not accessible (404 page). Confirm that add projects option is not available.
   * Hide news module menu item by settings menuItem to null. Confirm that news menu option is not shown but the news route can be accessed. Login as rsd admin and confim news option is available. Confirm add news option is available. On default RSD homepage, top 3 news items should be shown. In the rsd-admin news option should be enabled. 
   * Disable news module and confirm that add news option is not available.
* Change the order of menu items by changing order of the entries in the settings.json. For example place organisations at the first place and confirm that organisation is the first menu item.

PR Checklist:

*   [x] Update documentation
*   [x] Tests

### DIFF
--- a/documentation/docs/03-rsd-instance/02-configurations.md
+++ b/documentation/docs/03-rsd-instance/02-configurations.md
@@ -159,7 +159,6 @@ The `host` section of settings.json defines following settings. **Most of them s
 - `privacy_statement_url`: the link to your privacy statement page. Used on the user profile page to let user accept the privacy statement.
 - `software_highlights`: the definitions for the software highlights section on the top of the software overview page. You can specify title and the number of items loaded in the carousel. The default values are shown below.
 - `orcid_search`: should ORCID api be used when searching for contributors or team members? By default ORCID search api is enabled and the entries from RSD and ORCID are combined in the search result. If you want your users to be able to search only within the existing RSD entries set this value to false.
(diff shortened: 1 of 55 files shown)


jmaassen (approved):
Works as expected

PR #1607, opened 2025-08-19
Don't expose unpublished meta pages in API

### LINKED ISSUE #1604: Unpublished meta pages are accessible via API
Meta pages that were created, but not yet set to public, are publicly accessible via `/api/v1/meta_pages`. From an administrator's perspective, it is expected that the content of unpublished pages would not be accessible via the API either.

ewan-escience:

* Change row level security so only published meta pages are exposed in the API
* Rename the table from `meta_pages` to `meta_page`
* Patch upgrade of REST Assured

### How to test

* Sign in as admin
* Create an unpublished meta page (called "Public page" in the admin interface)
* It should not be shown at http://localhost/api/v1/meta_page when visiting in your browser
* Publish it
* It should now be present at the API
* Check if the new test does what it is supposed to do
* Check if the RLS is correct

closes #1604

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [x] Tests

### DIFF
--- a/backend-tests/README.md
+++ b/backend-tests/README.md
@@ -1,6 +1,6 @@
 <!--
-SPDX-FileCopyrightText: 2023 - 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-SPDX-FileCopyrightText: 2023 - 2024 Netherlands eScience Center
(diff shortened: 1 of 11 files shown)


jmaassen (approved):
Works as expected

PR #1615, opened 2025-09-03
Add categories to project metadata overview

### LINKED ISSUE #1596: add categories to the metadata overview
It would be very useful to have another column in the project metadata overview, which displays a number of categories for a given project. This way I could, for example, check if I added calls and infra information to the project pages.

ewan-escience:

* Add categories to project metadata overview, it counts assigned incidental leaf nodes (nodes that can have children but of which no child is assigned to the project)

### How to test

* Sign in as admin, look at the project metadata overview, the final columns should show the category count
* Create a new project, add an organisation with categories to it and assign one category to the project
* When saving the category, you will see that all parents are also saved in the dev tools of your browser (F12, network tab), but in the project metadata overview, it should be counted as one (this should also be the case whan assigning a category that does have children, but of which you don't assign any children)
* Check for other project pages if the category count makes sense

closes #1596

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/database/106-project-views.sql
+++ b/database/106-project-views.sql
@@ -312,6 +312,31 @@ GROUP BY
 	output_for_project.project;
 $$;
 
+
+-- only counts assigned incidental leaf nodes of the category tree
+-- (nodes that can have children but of which no child is assigned)
+CREATE FUNCTION count_project_categories() RETURNS TABLE (
+	project UUID,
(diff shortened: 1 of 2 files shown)


jmaassen (approved):
Works as expected

PR #1624, opened 2025-09-16
fix: obtain user data from Azure when signing in

### LINKED ISSUE #1620: No user information populated after connecting via Azure (openID)
After setting all the Azure-related details and successfully login for the first time, I need to accept the T&C and Privacy notice and then I can start using the tool. However, when I check the settings, I find them totally empty of any detail:

Image

I assume that this is because at no point we have pulled all of this information from Azure, but it is something that will be convenient to have if just for the user itself to have a meaningful account. 

I'm happy to take on this, but it would be great if your could point me in the right direction on where in the code this should be implemented.

ewan-escience:

* Obtain email and name from Azure when signing in by adding more scopes

### How to test

* Sign in with an Azure acocount
* Your name and email should be filled in in your profile

**Note**: This will only work for new RSD accounts, existing accounts will *not* be updated

closes #1620

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/authentication/src/main/java/nl/esciencecenter/rsd/authentication/AzureLogin.java
+++ b/authentication/src/main/java/nl/esciencecenter/rsd/authentication/AzureLogin.java
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


dmijatovic (approved):
Nice fix. It works with my Azure app.
@dalonsoa Can you confirm this fix will work for you too?

image

PR #1721, opened 2026-04-03
Fix releases layout and news images

### LINKED ISSUE #1546: Spacing between news images
When adding multiple news images, the spacing between images seems very large, especially for small images: 

Image

### LINKED ISSUE #1687: Version tags drop of the page in releases overview
When looking at the nlesc releases, I noticed some of the version tags drop of the page at the side:

Image

Would right aligning make more sense here?

ewan-escience:

* Right align release version tags so they don't extend too far to the right
* Fix spacing between small and tall news images

### How to test

* Sign in, add software, publish, add an organisation, add Vantage6 concept DOI: `10.5281/zenodo.7221216`
* `docker compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.doi.MainReleases`
* Check the releases page of the organisation, release tags should be right aligned (on of the releases has an unkown release date messing up the layout, but that's an independent issue that already existed)
* Create a news item, add some small and tall images, there should be no big gaps between the images
* Check if images in other places still render correctly

closes #1687
closes #1546

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/styles/global.css
+++ b/frontend/styles/global.css
@@ -7,6 +7,7 @@
  * SPDX-FileCopyrightText: 2023 - 2026 Netherlands eScience Center
  * SPDX-FileCopyrightText: 2025 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
  * SPDX-FileCopyrightText: 2026 Dusan Mijatovic (NLEsc) <d.mijatovic@esciencecenter.nl>
+ * SPDX-FileCopyrightText: 2026 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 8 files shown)


jmaassen (approved):
Works as advertised!

frontend/styles/global.css
```diff
@@ -115,7 +116,7 @@ body {
   */
   min-width: 30rem;
   /* fix jumping scrollbar */
-  overflow: overlay;
```
dmijatovic, line 118: Does jumping scrollbar fix still works after this change?
Maybe is better to use `overflow: auto` at the lower level where it fixes specific problem?

ewan-escience: > Does jumping scrollbar fix still works after this change?

I don't know. Do you have an example of where this happens?

I changed it because the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow) say:

> The keyword value overlay is a legacy value alias for auto. With overlay, the scroll bars are drawn on top of the content instead of taking up space.

PR #1741, opened 2026-04-30
Use OpenAlex mention types

### LINKED ISSUE #1723: Journal articles of reference papers and citations by CrossRef are mapped to "Other"
When adding a reference paper where the metadata is hosted at CrossRef, journal articles are classified as "Other" despite the correct annotation in the metadata. This also applies to all scraped citations.

Example: https://api.crossref.org/works/10.3389/fhpcp.2025.1709051

Shortened response:

```json
{
    "message": {
        "type": "journal-article",
   }
}
```

ewan-escience:

* Use the OpenAlex field `type` instead of the removed field `type_crossref`. The types can be found [here](https://developers.openalex.org/api-reference/work-types).
* Update the types from Crossref in accordance with [its types](https://api.crossref.org/types)
* Update the types from DataCite in accordance with [its types](https://schema.datacite.org/meta/kernel-4.7/)

### How to test

* Check the source code to see if you agree with the new mappings
* Sign in, add software
* Add reference papers, like `10.3389/fhpcp.2025.1709051`,  `10.1016/j.future.2024.07.050` or `10.1016/j.future.2018.08.004`
* `docker compose exec scrapers java -cp /usr/myjava/scrapers.jar nl.esciencecenter.rsd.scraper.doi.MainCitations`
* Most of the citations should not have mention type `Other`

closes #1723

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/utils/getCrossref.ts
+++ b/frontend/utils/getCrossref.ts
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: 2022 - 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-// SPDX-FileCopyrightText: 2022 - 2025 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2022 - 2026 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+// SPDX-FileCopyrightText: 2022 - 2026 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 dv4all
 // SPDX-FileCopyrightText: 2025 Dusan Mijatovic (Netherlands eScience Center)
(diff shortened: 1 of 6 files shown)


dmijatovic (approved):
It works for me with DOI mentioned. Tried few others too.
Maybe @jmaassen can find some specific ones that need more improvement?

jmaassen:
I noticed some issues with these DOIs: 

https://doi.org/10.1016/j.jpdc.2021.10.003
https://doi.org/10.1016/j.parco.2021.102813  

When I add them directly to an empty RSD, they are correctly identified (by crossref) as journal papers: 
Screenshot_2026-05-01_09-39-53

However, if I add this DOI (again to an empty RSD): 

https://doi.org/10.1016/j.future.2018.08.004

and run the scraper, it finds the to DOIs above on OpenAlex and classifies them as "other":

Screenshot_2026-05-01_09-43-38

(see the bottom and top one in the screenshot)

If I then try to add one of these papers as reference papers, it will no longer look at the correct crossref info, but just take the incorrect info from OpenAlex instead: 

Screenshot_2026-05-01_09-46-02

jmaassen:
If I look at the info in OpenAlex for 

https://doi.org/10.1016/j.parco.2021.102813 

I get this: 

https://api.openalex.org/w3082350385

Note that it is important here that the same record contains information for both the preprint on Arxiv and the actual published paper in the journal. We should take the type of the latter

ewan-escience:
We decided to merge this now, as we want to release this and this is an improvement over the current situation.

Note that citations will eventually be harvested by the regular mentions harvester, which will assign the right type. Those mentions are then not overwritten by the citations harvester again, so they keep their correct type. It's unfortunate that there thus is a time window in which a mention can have the wring type.

cmeessen:
Thanks for fixing. Will this fix already existing harvested mentions, or is there a strategy how to fix this manually?

ewan-escience:
Those should be fixed automatically by the regular mentions harvester.

PR #1744, opened 2026-05-07
48 no title

### LINKED ISSUE #1731: [A11y] 2.4.2 Page Titled - Page Title is Not Descriptive
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/48

dalonsoa:
# Pull request Title

Fixes https://github.com/research-software-directory/RSD-as-a-service/issues/1731

* Modifies the generation of metadata in the Edit Software and User pages such that the page title is informative about the contents of the page.
* Does the same for the software page, where a missing description was preventing the software title to correctly display.

How to test:

* Go to one of the above pages and check in the page source that the page title correctly represents what is being shown. For example, for the User `admin`, Settings > Profile shows shows `Profile | Settings | admin | Research Software Directory`

image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
+++ b/frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
@@ -1,3 +1,8 @@
+// SPDX-FileCopyrightText: 2026 Diego Alonso Alvarez (Imperial College London) <d.alonso-alvarez@imperial.ac.uk>
+// SPDX-FileCopyrightText: 2026 Imperial College London
+//
+// SPDX-License-Identifier: Apache-2.0
+
 import {Metadata} from 'next'
 import {notFound} from 'next/navigation'
 import EditSoftwareCommunities from '~/components/software/edit/communities'
@@ -17,9 +22,34 @@ import SoftwareRepositories from '~/components/software/edit/repositories'
 
 import {app} from '~/config/app'
 
-export const metadata: Metadata = {
-  title: `Edit software | ${app.title}`,
-  description: 'Edit software page',
+const editSoftwarePageTitle: Record<EditSoftwarePageId, string> = {
+  information: 'Information',
+  links: 'Links',
+  repositories: 'Repositories',
+  contributors: 'Contributors',
+  organisations: 'Organisations',
+  mentions: 'Mentions',
+  testimonials: 'Testimonials',
+  'package-managers': 'Package managers',
+  'software-heritage': 'Software Heritage',
+  communities: 'Communities',
+  'related-software': 'Related software',
+  'related-projects': 'Related projects',
+  maintainers: 'Maintainers',
+  services: ''
+}
+
+export async function generateMetadata({
+  params
+}:Readonly<{
+  params: Promise<{slug: string, page: EditSoftwarePageId}>
+}>): Promise<Metadata> {
+  const {page} = await params
+  const pageTitle = editSoftwarePageTitle[page]
+  return {
+    title: pageTitle ? `${pageTitle} | Edit software | ${app.title}` : `Edit software | ${app.title}`,
+    description: 'Edit software page',
+  }
 }
 
 export default async function EditSoftwarePageRouter({
--- a/frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
+++ b/frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
@@ -7,7 +7,7 @@ import {Metadata} from 'next'
 import {notFound} from 'next/navigation'
 import EditSoftwareCommunities from '~/components/software/edit/communities'
 import EditSoftwareContributors from '~/components/software/edit/contributors'
-import {EditSoftwarePageId} from '~/components/software/edit/editSoftwareMenuItems'
+import {editSoftwareMenuItems, EditSoftwarePageId} from '~/components/software/edit/editSoftwareMenuItems'
 import EditSoftwareDescriptionPage from '~/components/software/edit/information'
 import EditSoftwareLinksPage from '~/components/software/edit/links'
 import EditSoftwareMaintainers from '~/components/software/edit/maintainers'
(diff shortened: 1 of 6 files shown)


dalonsoa:
@dmijatovic I've not added tests because it is not clear to me how to do so. I have not been able to find where in the code the front end functionality is tested, or how. Happy to do so if you give me some guidance.

frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
```diff
@@ -17,9 +22,34 @@ import SoftwareRepositories from '~/components/software/edit/repositories'
...
+}:Readonly<{
+  params: Promise<{slug: string, page: EditSoftwarePageId}>
+}>): Promise<Metadata> {
+  const {page} = await params
+  const pageTitle = editSoftwarePageTitle[page]
+  return {
+    title: pageTitle ? `${pageTitle} | Edit software | ${app.title}` : `Edit software | ${app.title}`,
+    description: 'Edit software page',
+  }
 }
```
Comment (dmijatovic) [inline frontend/app/(base)/software/[slug]/edit/[page]/page.tsx:37]: We can solve it as proposed here with additional object but that would require additional maintaince. I would prefer to use label from the `editSoftwareMenuItems` item so we are always in sync with menu options/tabs.

Here is my proposal

```typescript
// in the header import
import {editSoftwareMenuItems, EditSoftwarePageId} from '~/components/software/edit/editSoftwareMenuItems'
 ...
  // find menuItem of current page we need to use in the page title metadata
  const menuItem = editSoftwareMenuItems.find(item => item.id === page)
  return {
   // use label of the menu item to added to page title
    title: menuItem ? `${menuItem.label} | Edit software | ${app.title}` : `Edit software | ${app.title}`,
    description: 'Edit software page',
  }
```

dmijatovic (changes requested):
I have one improvement suggestion you can start working on. In addition same approach need to be implemented in Edit project page. 

Concerning the User part I need more time to investigate. We have counts in the tab item labels there and I need to investigate how complex is to get these into metadata or we should modify `userTabItems` label to support labels without count instead of returning 0 or maybe add additional property `pageTitle`. I don't think that adding count to title is necessary. What do you think?

In addition, I noticed in User / Settings tabs when you click second time on the same tab the "correct" title will be removed/reset. This is high likely related to `useEffect` but I need more time to investigate that and confirm. 

I am at the conference tomorrow. I will have time to investigate further next Monday. Maybe you can implement suggested improvements in edit software and edit project pages in the meantime?

Thank you very much for your hard work.

dmijatovic:
> @dmijatovic I've not added tests because it is not clear to me how to do so. I have not been able to find where in the code the front end functionality is tested, or how. Happy to do so if you give me some guidance.

@dalonsoa, we have problems running frontend tests in CI/CD when contributions are done by external contributors. The frontend tests are not failing due to change in the code but because they cannot run under your account. At this moment we do not have solution for this problem. For now you can run unit test locally from the `frontend` folder using `npm run test` to see if the unit tests are really broken.

I will also run unit tests locally before approving your PR.

frontend/app/(container)/user/[tab]/page.tsx
```diff
@@ -1,14 +1,37 @@
...
+  const user = await getUserFromToken(token)
+  const {tab} = await params
+  const tabLabel = userPageTitles[tab] ?? 'User'
+
+  // if user exists we create metadata
+  return {
+    title: `${tabLabel} | ${user?.name ?? 'User'} | ${app.title}`,
+    description: `${tabLabel} pages for ${user?.name ?? 'User'}`
+  }
+}
```
Comment (dmijatovic) [inline frontend/app/(container)/user/[tab]/page.tsx]: generateMetadata function need to be extended with searchParams to handle the setting tabs.
Below is the proposed solution for this function.

```typescript
export async function generateMetadata({params,searchParams}: {
  params: Promise<{tab: UserPageId}>
  searchParams: Promise<{settings?: UserSettingsTab}>
}): Promise<Metadata> {
  // read route params
  const {token} = await getUserSettings()
  const user = await getUserFromToken(token)
  const [{tab},{settings}] = await Promise.all([
    params,
    searchParams
  ])
  const tabLabel = userPageTitles[tab] ?? 'User'

  // console.group('UserPageRouter.generateMetadata')
  // console.log('tab...', tab)
  // console.log('settings...', settings)
  // console.groupEnd()

  // when settings page we also need to extract settings tab
  if (tab==='settings'){
    // find settings tab
    const settingsTab = settingsMenu.find(item=>item.id===settings)
    const settingTitle = settingsTab?.label() ?? 'Profile'
    return {
      title: `${settingTitle} | ${user?.name ?? 'User'} | ${app.title}`,
      description: `${tabLabel} pages for ${user?.name ?? 'User'}`
    }
  }

  // other user pages without additional settings tab
  return {
    title: `${tabLabel} | ${user?.name ?? 'User'} | ${app.title}`,
    description: `${tabLabel} pages for ${user?.name ?? 'User'}`
  }
}
```

frontend/components/user/settings/UserSettingsContent.tsx
```diff
@@ -14,10 +17,22 @@ import UserProfilePage from './profile'
 import UserAgreementsPage from './agreements'
 import UserAccessTokensPage from './accesstokens'
 
+export const settingsSubpageTitles: Record<UserSettingsTab, string> = {
+  profile: 'Profile',
+  about: 'About Me',
+  agreements: 'User Agreements',
+  accesstokens: 'API Access Tokens',
+}
+
```
dmijatovic: Not needed when the proposed solution is implemented in `generateMetadata`

frontend/components/user/settings/UserSettingsContent.tsx
```diff
@@ -14,10 +17,22 @@ import UserProfilePage from './profile'
...
 export default function UserSettingsContent() {
   const searchParams = useSearchParams()
   const settings = searchParams?.get('settings') as UserSettingsTab ?? 'profile' as UserSettingsTab
 
+  useEffect(() => {
+    const subpageLabel = settingsSubpageTitles[settings] ?? 'Profile'
+    document.title = `${subpageLabel} | ${document.title}`
+  }, [settings])
+
```
dmijatovic: Not needed when proposed solution is implemented in `generateMetadata`. In addition this approach does not work properly when user clicks on the same tab twice. This is because this code "collides" with `generateMetadata` in the parent page.

frontend/components/user/tabs/UserTabItems.tsx
```diff
@@ -26,6 +28,16 @@ export type UserTabProps = {
...
+export const userPageTitles: Record<string, string> = {
+  software: 'My Software',
+  projects: 'My Projects',
+  organisations: 'My Organisations',
+  communities: 'My Communities',
+  'project-quality': 'Project metadata',
+  settings: 'Settings',
+  about: 'About',
+}
+
```
dmijatovic: If we decide to go with this approach then it would be more appropriate to type this as `Record<UserPageId, string>`. But let's first discuss if we want to have separate const or add `pageTitle` property to `userTabItems` object

dmijatovic (changes requested):
@dalonsoa These are my final improvement suggestions for User pages. Let me know if the suggestions are clear.

frontend/app/(base)/software/[slug]/edit/[page]/page.tsx
Comment (dalonsoa) [reply frontend/app/(base)/software/[slug]/edit/[page]/page.tsx:37]: This is way more elegant than the previous approach. Many thanks for suggesting.

frontend/app/(container)/user/[tab]/page.tsx
Comment (dalonsoa) [reply frontend/app/(container)/user/[tab]/page.tsx]: I tried this before, but I was not considering it a future and was always having the value `undefined`, so I tried the other option. This is, of course, better.

dalonsoa:
Changes implemented. Everything seems to be working and tests pass locally:

```
Test Suites: 155 passed, 155 total
Tests:       644 passed, 644 total
Snapshots:   0 total
Time:        8.973 s, estimated 9 s
```

dmijatovic:
@dalonsoa Great job! Similar changes should be done in project and high likely organisation and communities. Shell we merge this PR and create additional issue to address other modules? I can do that issue.

dmijatovic (approved):

dalonsoa:
I think it will be better to have separate issues for those, even if the solution is similar.

PR #1770, opened 2026-05-11
Fix double tab presses

### LINKED ISSUE #1729: [A11y] 2.4.3 Focus Order - Tab Order is Not Logical
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/50

ewan-escience:

* Keywords and other chips now have only one tab press instead of two
* Same for pagination items, like on the bottom of the software or project overview pages

### How to test

* Go to the software or project overview page
* The pagination buttons should only contain one tab press per button
* Clicking them should trigger client side navigation, not a full page reload
* Disable JavaScript, links should still work, but with a full page reload, and they don't have focus styling. We might improve on the styling when we upgrade to MUI v9, as the PaginationItem component [has a new `nativeButton` prop](https://mui.com/material-ui/api/pagination-item/#props).
* Visit a software or project page
* Keywords and categories (and research domains for projects) should only contain one tab press per button
* They should trigger client side navigation when JS is enabled
* They should still be clickable (except for categories) and have styling when JS is disabled
* Check if there are any other elements that still have double tab navigation

closes #1729

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/components/home/rsd/GradientBorderButton.tsx
+++ b/frontend/components/home/rsd/GradientBorderButton.tsx
@@ -1,13 +1,15 @@
 // SPDX-FileCopyrightText: 2024 - 2026 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2024 - 2026 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2026 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
 // SPDX-License-Identifier: Apache-2.0
 
+import Link from 'next/link'
+
 export default function GradientBorderButton({text, url, target = '_self', minWidth = '9rem'}: {text: string, url: string, target?: string, minWidth?: string}) {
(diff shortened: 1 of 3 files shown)


ewan-escience:
@dalonsoa can you also review this?

dalonsoa:
This looks really nice. Such a small change, and such a big impact. I cannot test it today, but will check as soon as possible, anyway.

dmijatovic (approved):
Nice work!

Let's wait for @dalonsoa to confirm that this works as required.

dalonsoa:
I've just tested this and it indeed seems to solve the issue. However, I've found in other parts of the page - not affecting Imperial's deployment - where that double tab is necessary to navigate. See for example:

[Screencast From 2026-05-13 07-07-53.webm](https://github.com/user-attachments/assets/1e669b58-4e9d-4383-afc3-e5426fb29805)

I do not know it that should be address as part of this PR or as a separate one.

ewan-escience:
Thanks for spotting this. I will try to fix it in this PR.

PR #1772, opened 2026-05-14
Add accessible name to feedback dialog

### LINKED ISSUE #1752: [A11y] 4.1.2 Name, Role, Value – ARIA Dialog Has No Accessible Name
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/66 for more details

dalonsoa:
# Add accessible name to feedback dialog

Fixes #1752 

Adds an accessible name to the feedback dialog. I believe this is what this issue required, although the accesiblity page pointed out does not contain the relevant information. The following one does, and that has been the approach taken.

https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

How to test:

* If you inspect the dialog now, it should have an accessible name. 

Screenshot From 2026-05-14 14-22-17

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/feedback/FeedbackPanelButton.tsx
+++ b/frontend/components/feedback/FeedbackPanelButton.tsx
@@ -95,9 +95,9 @@ User Agent: ${navigator.userAgent}`
       >
         <div className="h-full w-full bg-base-700 p-5 ">
           <div className="mx-auto max-w-[500px]">
-            <h1 className="text-base-100 mb-4" id="feedback-dialog-title">
+            <h2 className="text-base-100 mb-4" id="feedback-dialog-title">
               Send Feedback
-            </h1>
+            </h2>
             <textarea
               autoFocus
               className="placeholder:text-base-500 outline-0 p-2 w-full h-28 text-sm bg-base-700 text-base-100 border border-base-600 rounded-sm"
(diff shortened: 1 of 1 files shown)


frontend/components/feedback/FeedbackPanelButton.tsx
```diff
@@ -88,14 +90,14 @@ User Agent: ${navigator.userAgent}`
...
+        aria-labelledby="feedback-dialog-title"
         disableScrollLock={disable}
       >
         <div className="h-full w-full bg-base-700 p-5 ">
           <div className="mx-auto max-w-[500px]">
-            <div className="text-base-100 text-xl mb-4">
+            <h1 className="text-base-100 mb-4" id="feedback-dialog-title">
               Send Feedback
-            </div>
+            </h1>
```
dmijatovic: Please use h2 for this title as we already have h1 on the page(s)

dmijatovic (changes requested):
Nice work! I did double check it is prefered to use h2 as we already have h1 on the page.

frontend/components/feedback/FeedbackPanelButton.tsx
dalonsoa: I don't think it is a problem - the dialog is modal, so when it is open, the main page is not accessible and there's only one header, the one of the dialog. But happy to change, if still needed.

dmijatovic: Yes, please. After googling about the dialog/modal approach it is advised to use h2 if h1 is already on the page. I undestrand your reasoning but it seems that is more common to use h2.

dalonsoa: Sorted!

dmijatovic (approved):
Excellent. Thank you!

PR #1773, opened 2026-05-15
Make list items acceptable from the accessibility perspective

### LINKED ISSUE #1753: [A11y] 4.1.2 Name, Role, Value – Lists Incorrectly Structured
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/65 for more details

dalonsoa:
# Make list items acceptable from the accessibility perspective

Fixes #1753 

It seems that the `ListItemButton` control is creating an underlying html code for the menu items that is not accessible, as it does not convey the right information - they are not `<li>` objects, but `<a>`. Also, they have as role `button`, which apparently is not accessible, either. 

To fix this, I used the less customised `ListItem` and add the right functionality and similar styling manually. The result is a bit ugly and defeats the purpose of using the nice MUI components, so we might need to reconsider the approach.

How to test:

* Launch the tool and edit a software. Check the source code of the menu and see that the items now are `<li>` and have no `button` role. 
* Use a tool like `Accessibility Insights for Web` to make sure that particular issue is not an issue. 

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/edit/EditSoftwareNav.tsx
+++ b/frontend/components/software/edit/EditSoftwareNav.tsx
@@ -7,19 +7,19 @@
 // SPDX-FileCopyrightText: 2024 - 2025 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 // SPDX-FileCopyrightText: 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 // SPDX-FileCopyrightText: 2025 PERFACCT GmbH
+// SPDX-FileCopyrightText: 2026 Diego Alonso Alvarez (Imperial College London) <d.alonso-alvarez@imperial.ac.uk>
+// SPDX-FileCopyrightText: 2026 Imperial College London
 //
 // SPDX-License-Identifier: Apache-2.0
 
 'use client'
 import {useRouter, useParams} from 'next/navigation'
-import Link from 'next/link'
 
 import List from '@mui/material/List'
-import ListItemButton from '@mui/material/ListItemButton'
+import ListItem from '@mui/material/ListItem'
 import ListItemIcon from '@mui/material/ListItemIcon'
 import ListItemText from '@mui/material/ListItemText'
 
-import {editMenuItemButtonSx} from '~/config/menuItems'
 import useRsdSettings from '~/config/useRsdSettings'
 import {usePluginSlots} from '~/config/RsdPluginContext'
 import SvgFromString from '~/components/icons/SvgFromString'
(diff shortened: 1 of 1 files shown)


dalonsoa:
The tests are failing as I haven't bother to fix anything there until I get your feedback.

ewan-escience (changes requested):
The menu items are links, so they should be `a` elements. They can be `a` elements inside `li` elements though, to satisfy the accessibility concerns. Please change this (so the structure should become `<ListItem><Link>...</Link><ListItem>`), as now the menu items are not clickable like a true link anymore. An explicit `tabIndex` is then also not required anymore.

This should then also be changed on other edit pages, like projects, but we can do that.

dmijatovic:
@dalonsoa and @ewan-escience The problem is a list without li items. So we can solve this in few ways in my opnion:

1. As suggested now by Diego, this removes link "feature" from the menu and the user will not be able to open link in new tab. But it is correct from html structure and A11y perspective.
2. As suggested by Ewan using List item and Link components to achieve nav>ul>li structure and have a link features.
3. I think we can remove `List` component (that is ul part) and just use `ListItemButton` as a link. The html structure will be in this case `nav > a`, which is syntactically correct and should work with A11y. If this works properly with A11y (for me it look like it works) this amounts to least changes to current code (and other menu's we need to update - eg. edit projects). 

I am interested in your thoughts on this 3 options.

ewan-escience:
Option 1 is not an option for me. Semantically the items are links, so they should be `a` tags.

Option 3 still involves a button I think, so you're fighting MUI then to remove all button artifacts.

I think option 2 is best. We could also remove the list and only have `nav > a`. This is up for debate [here](https://css-tricks.com/navigation-in-lists-to-be-or-not-to-be/) and [here](https://www.sitepoint.com/are-navigation-lists-necessary/).

dmijatovic:
OK. Option 2 is most complete, A11y proof and feature reach approach `nav>ul>li>a`. I think we can achieve it by wrapping ListItemButton in ListItem. I'll have a look at it and come with proposed approach.

ewan-escience:
Can you skip `ListItemButton`? You'll probably have to deal with `role=button` and maybe other artifacts.

dmijatovic:
@ewan-escience @dalonsoa I have created new PR #1776  that addesses list items of edit software and edit project menu items. I did not encounter role=button here but you should validate this approach.

dalonsoa:
After https://github.com/research-software-directory/RSD-as-a-service/pull/1776, this one can be closed.

PR #1776, opened 2026-05-15
1753 list structure

### LINKED ISSUE #1753: [A11y] 4.1.2 Name, Role, Value – Lists Incorrectly Structured

dmijatovic:
# Improve edit software and edit project navigation for a11y

Closes #1753 
Closes #1773 

* Wrap edit software menu items with ListItem in order to achieve semantically correct structure nav>ul>li>a
* Add a11y aria markup to communicate menu items, current menu item and hide icons
* Improve global search markup for a11 based on  `Accessibility Insights for Web` chrome plugin feedback

How to test:
* login as rsd admin to be able edit software and project
* navigate to edit software pages
* confirm menu structure is: nav > ul > li > a 
* use browser plugin `Accessibility Insights for Web` to confirm A11y of the menu items
* perform same check for edit project menu items
* perform a11y check on global search (it's same on all pages)

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/GlobalSearchAutocomplete/index.tsx
+++ b/frontend/components/GlobalSearchAutocomplete/index.tsx
@@ -294,7 +294,7 @@ export default function GlobalSearchAutocomplete(props: Props) {
               fill="#707070"/>
           </svg>
         </div>
-        <label
+        <section
           role="search"
           aria-label="Search RSD website"
         >
@@ -323,14 +323,14 @@ export default function GlobalSearchAutocomplete(props: Props) {
             // Tells which item is highlighted via arrow keys
             aria-activedescendant={selected !== -1 && isOpen ? `search-item-${selected}` : undefined}
           />
-        </label>
+        </section>
         <span
           className="absolute top-[9px] right-2 text-base-600"
           hidden={isOpen}
         >{searchCombo}</span>
 
         {isOpen &&
-          <ul // NOSONAR
+          <ul
             // AI-assisted suggestion Google seach AI-Mode
             // Matches the aria-controls tag on the input element
             id="global-search-results"
@@ -359,14 +359,14 @@ export default function GlobalSearchAutocomplete(props: Props) {
               })
 
               return (
-                <li // NOSONAR
-                  key={item.slug}
+                <li
+                  key={url}
                   // Distinct ID mapped to aria-activedescendant
                   id={`search-item-${index}`}
-                  // AI-assisted suggestion Google seach AI-Mode
+                  // AI-assisted suggestion Google search AI-Mode
                   // Explicitly states this element is a selectable list option
                   role="option"
-                  // AI-assisted suggestion Google seach AI-Mode
+                  // AI-assisted suggestion Google search AI-Mode
                   // Dynamic flag tracking current keyboard focus
                   aria-selected={selected === index}
                   data-testid="global-search-list-item"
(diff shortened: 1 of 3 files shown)


ewan-escience (changes requested):
The lists seem to work well.

For the global search bar, I'm still getting warnings in my broswer about the missing label for the input and the image.

- For the search bar input, you can add the `id` and `aria-labelledby` (having the same value) to the `input` element. 
- For the search icon, I'm not whether it's better to hide it or to add a label.

dalonsoa:
I cannot check this today (will do first thing tomorrow) but the list looks correct to me. I agree that for the search bar we will need an aria-labelledby and id, as it was being done in other places where we are facing the same problem.

dmijatovic:
@ewan-escience and @dalonsoa thanks for your feedback. I did more research specific to global search component using bit of AI assistance. Additional improvement suggestions are made to global search to make it more accessible.  Can you validate this please?

Concerning `aria-labelledby` approach, I want to note that this is equivalent to `aria-label` approach with a difference that when input has a label then `aria-labelledby` is preferred/used and when there is no label the `aria-label` can be used directly on the component.

ewan-escience:
> I did more research specific to global search component using bit of AI assistance.

According to our [internal policy on AI](https://doi.org/10.5281/zenodo.10363728):

> to ensure full transparency, and in support of the principles of open science, it is obligatory to
explicitly indicate which parts of your output have been AI-assisted and/or AI-generated, and by
which Generative AI tool (incl. version number, if available and relevant); although proper
legislation is still lacking, failing to do so may be an offense similar to, or bordering on,
plagiarism or fraud.

So can you include that?

dmijatovic:
@ewan-escience Let me know if this is clear

frontend/components/GlobalSearchAutocomplete/index.tsx
```diff
@@ -277,35 +278,69 @@ export default function GlobalSearchAutocomplete(props: Props) {
...
-          placeholder="Search or jump to..."
-          autoComplete="off"
-          value={inputValue}
-          onChange={handleChange}
-          onKeyDown={handleKeyDown}
-          type="text"
-          onFocus={focusSearch}
-          aria-label="Search RSD website"
+        <label
           role="search"
```
dalonsoa, line 297: I'm getting a complain that `ARIA role search is not allowed for given element`, as it can only be assigned to focusable elements. I suggest you move this role to the `input` element below (as it was originally).

dalonsoa (changes requested):
Aside from the note below, the list part does not report any accessibility error anymore.

dalonsoa (approved):
This works!

ewan-escience:
The FIrefox devtools still say that the global search input should [have a _visible_ label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Guides/Understanding_WCAG/Text_labels_and_names#form_elements_should_have_a_visible_text_label).

And can you merge your license entries (with and without email)?

dmijatovic:
@ewan-escience I fixed the license log. Concerning the label, I think in this case the label does not apply. The design layout does not "calls" for label. For a11y we implemented all required features. Placing the label will break current design, adding a label and then hiding it to achieve a11y functionality is not required as we already handle it. 

The current approach covers all functional requirements while respecting the current design requirement. Let me know if you think otherwise.

ewan-escience:
Well, I think Mozilla's documentation is clear enough. We could, however, decide that it's good enough (in terms of accessibility) for now, as indeed a redesign is needed otherwise. If someone with a screenreader reports to us that it is not good enough, we should redesign it.

ewan-escience (approved):

PR #1786, opened 2026-05-21
Add description to commit chart

### LINKED ISSUE #1748: [A11y] 1.1.1 Non-Text Content - Complex Images Do Not Have Equivalent Alternative Text
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/58 for more details

dalonsoa:
# Add description to commit chart

Fixes #1748 

* Add a description to the commit history chart, to inform the screen readers users of its content, indicating the start and end date of the history. Describing the individual data points is probably not practical.
* Also add `role=img`, so the object can correctly be identified as an image.
* This follows the recommendations indicated by [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role#svg_and_roleimg).  

How to test:

* Run the tool with some test data with `make start`
* In a software with a commit history, the graph should show an aria-label and its role as an image. Something like the following:
* 
Screenshot From 2026-05-21 06-42-02

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/software/CommitsChart.tsx
+++ b/frontend/components/software/CommitsChart.tsx
@@ -75,9 +75,8 @@ export default function CommitsChart({
 
     // Get the date range of the commits to build a meaningful description
     // define date formatting using locale: default = user specific, en_GB = if user local fails etc.
-    const locale = ['default','en-GB','en-US']
-    const fist_date = new Date(lineData[0].x).toLocaleDateString(locale)
-    const last_date = new Date(lineData[lineData.length - 1].x).toLocaleDateString(locale)
+    const fist_date = new Date(lineData[0].x).toLocaleDateString()
+    const last_date = new Date(lineData[lineData.length - 1].x).toLocaleDateString()
     const description = `Curve showing the daily commits during the history of the software, from ${fist_date} to ${last_date}.`
 
     // render
--- a/frontend/components/software/CommitsChart.tsx
+++ b/frontend/components/software/CommitsChart.tsx
@@ -5,6 +5,8 @@
 // SPDX-FileCopyrightText: 2023 Dusan Mijatovic (dv4all) (dv4all)
 // SPDX-FileCopyrightText: 2024 - 2025 Dusan Mijatovic (Netherlands eScience Center)
 // SPDX-FileCopyrightText: 2024 - 2025 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2026 Diego Alonso Alvarez (Imperial College London) <d.alonso-alvarez@imperial.ac.uk>
+// SPDX-FileCopyrightText: 2026 Imperial College London
 //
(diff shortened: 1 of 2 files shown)


frontend/components/software/CommitsChart.tsx
```diff
@@ -70,10 +72,20 @@ export default function CommitsChart({
   if (commit_history && Object.keys(commit_history).length > 0) {
     // format commits data for chart and calculate other stats
     const {lineData, lastCommitDate, totalCountY} = prepareDataForSoftwarePage(commit_history)
+
+    // Get the date range of the commits to build a meaningful description
```
ewan-escience, line 76: The date format is now hard coded. I would prefer to show it in the locale of the user, using [`toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString).

ewan-escience (changes requested):
Works well, but I've got one request regarding the formatting of the date.

frontend/components/software/CommitsChart.tsx
```diff
@@ -70,10 +72,20 @@ export default function CommitsChart({
...
+    let date = new Date(lineData[0].x)
+    const fist_date = date.getDate() + '/' + (date.getMonth() + 1) + '/' +
+    date.getFullYear()
+    date = new Date(lineData[lineData.length - 1].x)
+    const last_date = date.getDate() + '/' + (date.getMonth() + 1) + '/' +
+    date.getFullYear()
+    const description = `Curve showing the daily commits during the history of the software, from ${fist_date} to ${last_date}.`
```
dmijatovic: Agree with Ewan. here is the suggestion

```typescrpt
// Get the date range of the commits to build a meaningful description
// define date formatting using locale: default = user specific, en_GB = if user local fails etc.
const locale = ['default','en-GB','en-US']
const fist_date = new Date(lineData[0].x).toLocaleDateString(locale)
const last_date = new Date(lineData[lineData.length - 1].x).toLocaleDateString(locale)
const description = `Curve showing the daily commits during the history of the software, from ${fist_date} to ${last_date}.`
```

dmijatovic (changes requested):
I made improvement suggestion.

dalonsoa:
Yes, I totally agree with this. Will make the change.

dmijatovic:
> Yes, I totally agree with this. Will make the change.

@dalonsoa Maybe just use @ewan-escience suggestion without locale "fallback" array I suggested. I realized that local fallback options should be approached from the global perspective (not only in this chart).

dmijatovic (approved):
Excellent work! Sorry for confusion about locale fall-back

ewan-escience (approved):

PR #1789, opened 2026-05-22
1730 fieldset text field

### LINKED ISSUE #1730: [A11y] 1.3.1 Info and Relationships - Related Form Elements Not Grouped Correctly
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/62

dmijatovic:
# Fieldset in text field

Closes #1730

@dalonsoa  I have remarks concerning this and the original issue. If I read the improvement suggestion correctly, it marks use of `<fieldset>` element in the material-ui TextField component as confusing for screen readers. The `fieldset` is used as visual element in "outlined" text type to draw outline around the text box. Our design called specifically for the outlined type as "default" type was somewhat "unclear" to our design specialist. Although `<fieldset>` element is used it is disabled for the screen readers using `aria-hidden="true"`. The screen readers should ignore this `fieldset` and the child element(s). **Can you validate this with the audit team? If this is correct assumption please approve this PR, otherwise we would need to consider alternative approaches:  using default type or developing custom component.**

* No changes are performed on text field. The fieldset is used as style item and it is disabled for screen readers by `aria-hidden="true"`.
* The element id is made more robust. This was not flagged as problem but based on the fixes performed on select component the improvement on element id is implemented.
* The contrast of avatar background is improved. The custom background color is removed which had not sufficient contrast.

How to test:
* login as user
* navigate to user section and run `Accessibility Insights For Web` plugin or other a11y tool. There should be no a11y warnings or errors.

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/e2e/helpers/project.ts
+++ b/e2e/helpers/project.ts
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
-// SPDX-FileCopyrightText: 2023 - 2025 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2023 - 2025 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2023 - 2026 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2023 - 2026 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -99,7 +99,7 @@ export async function createProjectLink(page: Page, {label, url}: { label: strin
   // add link label in the modal
   await page.getByRole('dialog', {
     name: 'Project link'
-  }).locator('#title').fill(label)
+  }).getByLabel('Title').fill(label)
 
   // add url
   await page.getByLabel('Url').fill(url)
--- a/e2e/helpers/utils.ts
+++ b/e2e/helpers/utils.ts
@@ -1,8 +1,8 @@
 // SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all)
 // SPDX-FileCopyrightText: 2022 - 2023 Dusan Mijatovic (dv4all) (dv4all)
 // SPDX-FileCopyrightText: 2022 - 2023 dv4all
-// SPDX-FileCopyrightText: 2023 - 2024 Dusan Mijatovic (Netherlands eScience Center)
-// SPDX-FileCopyrightText: 2023 - 2025 Netherlands eScience Center
+// SPDX-FileCopyrightText: 2023 - 2026 Dusan Mijatovic (Netherlands eScience Center)
+// SPDX-FileCopyrightText: 2023 - 2026 Netherlands eScience Center
 // SPDX-FileCopyrightText: 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 //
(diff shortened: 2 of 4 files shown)


dalonsoa:
@dmijatovic I just forwarded the query to the testers. Let's say what they say. In my opinion, your solution should be alright - as it stands, the fieldsets should not be a problem for screen readers. I'll let you know what they answer, anyway.

dalonsoa:
@dmijatovic got an answer:

> If an element has aria-hidden="true", any interactive content inside it becomes a 'silent focus trap'. Screen reader users can still tab into the fields, but their assistive technology will remain completely silent.

It seems there's some work for us to do.

dmijatovic:
@dalonsoa I need to push back a bit on this issue. I have further investigated it. I am not expert in a11y and I cannot test with the real screen reader. Your feedback mentions `any interactive content inside it becomes a 'silent focus trap'`. However, there is no any interactive content inside the `<fieldset>` in this case as far as I can see. 

Below is rendered/raw html of one of TextField components:

```html
<div class="MuiFormControl-root MuiFormControl-fullWidth MuiTextField-root mui-cmpglg-MuiFormControl-root-MuiTextField-root">
  <label
    class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-outlined MuiFormLabel-colorPrimary MuiFormLabel-filled MuiInputLabel-root MuiInputLabel-formControl MuiInputLabel-animated MuiInputLabel-shrink MuiInputLabel-outlined mui-zunwc2-MuiFormLabel-root-MuiInputLabel-root"
    data-shrink="true"
    for="_r_1u_-given_names"
    id="_r_1u_-given_names-label"
  >
    First name / Given name(s)
  </label>
  <div class="MuiInputBase-root MuiOutlinedInput-root MuiInputBase-colorPrimary MuiInputBase-fullWidth MuiInputBase-formControl mui-180qf2u-MuiInputBase-root-MuiOutlinedInput-root">
    <input
      id="_r_1u_-given_names"
      type="text"
      aria-invalid="false"
      aria-describedby="_r_1u_-helper-text"
      autocomplete="off"
      class="MuiInputBase-input MuiOutlinedInput-input mui-2tcmkv-MuiInputBase-input-MuiOutlinedInput-input"
      value="Jordan"
    >
    <fieldset
      aria-hidden="true"
      class="MuiOutlinedInput-notchedOutline mui-18p5xg2-MuiNotchedOutlined-root-MuiOutlinedInput-notchedOutline">
      <legend class="mui-ex8a5f-MuiNotchedOutlined-root">
        <span>First name / Given name(s)</span>
      </legend>
    </fieldset>
  </div>
  <p
    class="MuiFormHelperText-root MuiFormHelperText-sizeMedium MuiFormHelperText-contained MuiFormHelperText-filled mui-13uvi2e-MuiFormHelperText-root"
    id="_r_1u_-helper-text"
  >
    <span class="mr-2">One or more given names</span><span>6/200</span>
  </p>
</div>
```

This issue has been discussed on the material-ui repository multiple times.  It seems that [material-ui team defends use of `<fieldset>` in this case](https://github.com/mui/material-ui/issues/23389). 

I propose to double check my feedback with the audit team and I would like to postpone decision after we have more in depth discussion where we can discuss the options?

dalonsoa:
I agree. I'll get back to them with your feedback and see where we land. I suspect their comment is based on an observation of the underlying code, and not actually the result of finding issues with the screen reader. If we can confirm to them that all uses of fieldsets are either the traditional usage - to indicate a set of fields related to each other - or in this way, not enclosing any interactive elements that could be masked, I think that will keep them happy.

dalonsoa:
I've raised this with the accessibility testing team. I'll keep you posted.

dmijatovic:
@dalonsoa I have closed this issue "accidentally" because I had this code in the PR that was approved. However, if the current approach is not sufficient we will need to re-open it or create new issue. Sorry about this.

PR #1793, opened 2026-05-26
Fix parsing name

### LINKED ISSUE #1676: Reversed name and surname in Azure
When a new user login using Azure, their profile seems to be populated assigning the name to the surname and vice versa. Specifically, we have the following cases when they check their profile page after first login, but it has been reported by other users, as well.

> At the top it says "Cummins, Daniel", but the First Name / Given Name(s) field says "Cummins," and the Last Name / Family Name(s) field says "Daniel"

> The name does not appear in the correct order for me - it should be the other way around. Also part of the name is truncated to only have the initials.
>
> Right now it says:
> First name/ Given name(s): "Bhogal,"
> Last name/ Family name(s): "Saranjeet Kaur S S"

Not sure if it is an issue in how the RSD handles the information sent by Azure or if Azure itself is sending things in the wrong order.

ewan-escience:

* When the OpenID name contains `, `, assume the family names come first when creating a public profile

### How to test

* Sign in with an OpenID provider that returns your name in the form `FamilyName, FirstName` and check that your name in your profile is correct
* If this is not possible for you, check the code and the tests

closes #1676

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [x] Tests

### DIFF
--- a/frontend/sonar-project.properties
+++ b/frontend/sonar-project.properties
@@ -1,6 +1,6 @@
+# SPDX-FileCopyrightText: 2023 - 2026 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
+# SPDX-FileCopyrightText: 2023 - 2026 Netherlands eScience Center
 # SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
Looks good to me.

dmijatovic (approved):
This might not be most robust solution but we can use it as quick fix.

Issue #1769, opened 2026-05-07
[A11y] 1.4.3 Contrast (Minimum) - Copy Has Insufficient Contrast Against Background


dalonsoa:
See https://github.com/ImperialCollegeLondon/RSD-as-a-service/issues/37 for details.

This is on us, but I add it here for completeness as it might affect the pages of other people.

dalonsoa:
Mostly sorted, but still some areas of improvement. Will work on them over the coming weeks. It has minor impact on accessibility, it would seem, so it should be OK.

PR #1796, opened 2026-05-28
Increase contrast of several aspects of the web

dalonsoa:
# Increase contrast of several aspects of the web

Fixes partly #1769 . No closing it, for now.

* There are several parts of the web where contrast is not good enough, or where we are conveying information only by colours. Part of it is just changing the colour palette - deployment specific - but other things require changes in the code itself. For example, there are places where the colours are hardcoded instead of using the palette. The feedback form is specially challenging as there, the `primary` colour needs to have good enough contrast against a dark background and as a background itself for a clear font. I have not been able to solve that for Imperial palette, yet. 

How to test:

* Run `docker compose up`. We do not want content in some places as the changes here concern some placeholder text.
* Visit the `feedback` dialog and check the contrast. With the standard palette, proably either the send feedback button or open a new issue link at the bottom will raise an error when checked with the accessibility tool. This is to be fixed, in principle, by playing with the colours of the palette.
* Go to the software of projects page. There should be no complain about the background text and image saying there's no content.
* Create a new software and go to the edit page. The Click or drop to upload your logo placeholder should raise no issues concerning contrast, either. 
* I have fixed a few other issues that were purely related with the Imperial palette, but the above should be helpful for all cases. 

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/frontend/components/layout/ImageWithPlaceholder.tsx
+++ b/frontend/components/layout/ImageWithPlaceholder.tsx
@@ -41,7 +41,7 @@ export default function ImageWithPlaceholder({
     }
     return (
       <div
-        className={`flex flex-col justify-center items-center text-base-600 rounded-xs ${className ?? ''}`}
+        className={`flex flex-col justify-center items-center text-base-700 rounded-xs ${className ?? ''}`}
       >
         <PhotoSizeSelectActualOutlinedIcon
           sx={{
--- a/frontend/components/layout/ImageWithPlaceholder.tsx
+++ b/frontend/components/layout/ImageWithPlaceholder.tsx
@@ -41,7 +41,7 @@ export default function ImageWithPlaceholder({
     }
     return (
       <div
-        className={`flex flex-col justify-center items-center text-base-700 rounded-xs ${className ?? ''}`}
+        className={`flex flex-col justify-center items-center text-base-600 rounded-xs ${className ?? ''}`}
       >
         <PhotoSizeSelectActualOutlinedIcon
           sx={{
--- a/frontend/components/layout/NoContent.tsx
+++ b/frontend/components/layout/NoContent.tsx
@@ -28,7 +28,7 @@ const NoContentBody = styled('div')({
   flexDirection: 'column',
   justifyContent: 'center',
   alignItems: 'center',
-  opacity: 0.50
+  opacity: 0.5
 })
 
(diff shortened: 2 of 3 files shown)


dalonsoa:
This is the updated feedback form for Imperial. The link at the bottom does not have enough contrast, but making the blue clearer, makes the button not having enough contrast. We do not want that colour (`primary`) to be too clear as it is used in other places over a clear background. My suggestion would be to use another colour - not `primary` for the link at the bottom. Maybe `accent` or something like that. 

image

frontend/components/layout/ImageWithPlaceholder.tsx
```diff
@@ -41,7 +41,7 @@ export default function ImageWithPlaceholder({
     }
     return (
       <div
-        className={`flex flex-col justify-center items-center text-base-600 rounded-xs ${className ?? ''}`}
+        className={`flex flex-col justify-center items-center text-base-700 rounded-xs ${className ?? ''}`}
```
dmijatovic: I already increased this one in previous PR's. Text-base-600 already solved the contrast issue with our color scheme. Can you validate this and revert back?

frontend/components/layout/NoContent.tsx
```diff
@@ -26,7 +28,7 @@ const NoContentBody = styled('div')({
   flexDirection: 'column',
   justifyContent: 'center',
   alignItems: 'center',
-  opacity: 0.25
+  opacity: 0.50
```
dmijatovic: SonarQube prefers 0.5. Please improve

dmijatovic (approved):
Nice work. I do have 2 minor remarks.

frontend/components/layout/NoContent.tsx
dalonsoa: Done

frontend/components/layout/ImageWithPlaceholder.tsx
dalonsoa: It works. I had an slightly different 600 grey in the Imperial palette, but with the standard collection of greys, it works.

dmijatovic:
@dalonsoa Thanks! I believe you can merge this PR's yourself but if you like me to do let me know. We require at least one approval before merge button is enabled. In some cases we discuss who will be the "main" person to approve PR but in this case one approval is sufficient.

PR #1797, opened 2026-05-28
feat: split frontend checks

### LINKED ISSUE #1781: Update github action dependencies
Release action and high likely other action need upgrades.

Also consider splitting unit tests and SonarQube into separate actions. It seems that SonarCube fails when PR's are made from forks. The unit test should run without problems so it is better to have these as separate actions to be clear what part fails.

dmijatovic:
# Split frontend checks and update actions

Close #1781

* Split frontend checks in 4 separate checks to directly see where is the problem. The checks are fe-lint, fe-test, fe-build, fe-sonar.  
* Sonar actions are upgraded to latest versions
* Release action dependencies upgraded to avoid Node version warnings

How to test:
* This can be only tested by creating PR with some errors.
* I tested it by purposely creating lint warning which should be signaled in checks
* I saw the warning from SonarCube too, similair to linter warning
* I tested build and unit test failures.
* **Release action upgrades can be only tested by making a release. We need to factor this in by our next release**. 

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.github/workflows/_cff.yml
+++ b/.github/workflows/_cff.yml
@@ -1,8 +1,8 @@
 # SPDX-FileCopyrightText: 2022 Dusan Mijatovic (dv4all)
 # SPDX-FileCopyrightText: 2022 dv4all
+# SPDX-FileCopyrightText: 2024 - 2026 Dusan Mijatovic (Netherlands eScience Center)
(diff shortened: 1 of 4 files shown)


ewan-escience (approved):
Looks good!

PR #1800, opened 2026-05-29
Ensure sonarqube action can be runned by fork PR's

dmijatovic:
# SonarQube and PR's from fork repos

* Split frontend test and sonar qube action in order to run sonarqube within local repo environment
* Sonar is separate action and runs after all other checks are done passed successfully.

How to test:
* This PR should be used for testing. All tests should run including SonarQube which will run after frontend tests.
* I tested linter failure/warnings
* I tested build failure
* I tested unit test failure
* The sonarqube action can be only tested after reaching main branch. **GitHub imposes a strict security restriction on the workflow_run event trigger: The workflow file must exist on the default branch (main) before it can ever be triggered. In order to test this change completely we need to merge this PR and then use next frontend PR to confirm that SonarQube action works properly.**
 

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.github/workflows/frontend_sonar.yml
+++ b/.github/workflows/frontend_sonar.yml
@@ -35,19 +35,7 @@ jobs:
           # Isolates the code string inside a hard token context to block injection tricks
           HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
 
-      # 2. Re-create the folder structure
-      - name: "install node and cache npm"
-        uses: actions/setup-node@v6
-        with:
-          node-version: 24.14
-          cache: 'npm'
-
-      - name: "install dependencies"
-        run: |
-          cd frontend
-          npm ci
-
-      # 3. DOWNLOAD ARTIFACT CROSS-WORKFLOW (test-coverage)
+      # 2. DOWNLOAD ARTIFACT CROSS-WORKFLOW (test-coverage)
       - name: "download coverage artifact"
         uses: actions/download-artifact@v7
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/frontend_sonar.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
+          # Use the safe string variable below
+          ref: ${{ env.HEAD_BRANCH }}
+          fetch-depth: 0
+        env:
+          # Isolates the code string inside a hard token context to block injection tricks
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+
+      # 2. Re-create the folder structure
```
ewan-escience: Why is this step needed? This looks dangerous, as it installs code which could contain malicious `pre-install` scripts. Furthermore, the result of the installation process doesn't seem to be used later on. Please either remove this step or document why it is safe _and_ necessary.

.github/workflows/frontend_sonar.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: "download coverage artifact"
+        uses: actions/download-artifact@v7
+        with:
+          name: frontend-coverage
+          path: frontend/coverage/
+          # Explicitly targets the parent workflow run to fetch its files
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # 4. Cache Sonar packages natively
```
ewan-escience: I'm also not sure why this step is needed. I don't think we install any packages from Sonar as we just use the action in the next step.

ewan-escience (changes requested):
The overall setup looks good, but I think there are some unnecessary steps that should be removed.

dmijatovic:
@ewan-escience Thanks for noticing npm section. Thisone is not required. However the caching of SonarQube is beneficial. As fas as I can find the SonarCube runner is copied into PR to be able to run. It also benefits "history" when multiple pushed are done on same PR. So I am for keeping the caching part.

ewan-escience (approved):

PR #1807, opened 2026-06-03
Use PostgREST ready flag

### LINKED ISSUE #1805: Use PostgREST ready flag insead of `wait-on`
We should use PostgREST's new [ready flag](https://docs.postgrest.org/en/v14/references/cli.html#ready-flag) to check when the backend is available for the data generation script and in the Docker Compose `healthcheck` to mark the container as healthy.

ewan-escience:

* Use PostgREST's new [ready flag](https://docs.postgrest.org/en/v14/references/cli.html#ready-flag) to mark the backend container as healthy
* Both the data-generation and the backend-test module now use this healthcheck to determine when they should start (`wait-on` for the data-generatio module was removed)
* Upgrade Faker from `10.3.0` to `10.4.0` ([release notes](https://github.com/faker-js/faker/releases))
* Upgrade PostgREST from `v14.6` to `v14.12` ([release notes](https://github.com/PostgREST/postgrest/releases))

### How to test

* **Add the new env variable from `.env.example` to your `.env` and `.env.local`**
* The data-generation script should start within one second of the backend container being ready
* Not sure why, but the data-generation script should run much faster than before (even in the commit before PostgREST was upgraded). UPDATE: this is probably because `wait-on` was making a request to the slow Swagger endpoint four times per second, which was taking up the connections from the connection pool.
* Check if the RSD still works
* Stop the containers and remove them (`docker compose down --volumes`)
* Check if the backend tests still pass (follow the instructions from `backend-tests/README.md`)
* Carefully check if the changes to the Docker Compose files (especially `deployment/docker-compose.yml`) are correct and if I didn't forget any changes in these or other files

closes #1805

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,7 +51,9 @@ services:
       - PGRST_DB_URI=postgres://rsd_authenticator:${POSTGRES_AUTHENTICATOR_PASSWORD}@${POSTGRES_DB_HOST}:${POSTGRES_DB_HOST_PORT}/${POSTGRES_DB}
       - PGRST_DB_ANON_ROLE
       - PGRST_DB_SCHEMA
+      - PGRST_SERVER_HOST=backend
       - PGRST_SERVER_PORT
+      - PGRST_ADMIN_SERVER_PORT
       - PGRST_JWT_SECRET
       - PGRST_DB_POOL=30
       - PGRST_DB_POOL_ACQUISITION_TIMEOUT=30
@@ -59,6 +61,13 @@ services:
       - database
     networks:
       - net
+    healthcheck:
+      test: [ "CMD", "postgrest", "--ready" ]
+      interval: 5s
+      retries: 2
+      timeout: 5s
+      start_period: 60s
+      start_interval: 1s
 
   auth:
     build: ./authentication
@@ -326,7 +335,8 @@ services:
       - PGRST_JWT_SECRET
       - POSTGREST_URL
     depends_on:
-      - backend
+      backend:
+        condition: service_healthy
     networks:
       - net
     deploy:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,7 +58,8 @@ services:
       - PGRST_DB_POOL=30
       - PGRST_DB_POOL_ACQUISITION_TIMEOUT=30
     depends_on:
-      - database
+      database:
+        condition: service_healthy
     networks:
       - net
     healthcheck:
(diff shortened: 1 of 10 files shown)


docker-compose.yml
```diff
@@ -51,14 +51,23 @@ services:
...
       - PGRST_DB_ANON_ROLE
       - PGRST_DB_SCHEMA
+      - PGRST_SERVER_HOST=backend
       - PGRST_SERVER_PORT
+      - PGRST_ADMIN_SERVER_PORT
       - PGRST_JWT_SECRET
       - PGRST_DB_POOL=30
       - PGRST_DB_POOL_ACQUISITION_TIMEOUT=30
     depends_on:
       - database
```
dmijatovic: Maybe add explicit healthy dependency for database here too as done in publisher service?

```yml
...
depends_on:
     database:
         condition: service_healthy
...
```

I think other services like `auth` and `background-service` might also have improved healthy checks.

docker-compose.yml
```diff
@@ -51,14 +51,23 @@ services:
       - PGRST_DB_URI=postgres://rsd_authenticator:${POSTGRES_AUTHENTICATOR_PASSWORD}@${POSTGRES_DB_HOST}:${POSTGRES_DB_HOST_PORT}/${POSTGRES_DB}
       - PGRST_DB_ANON_ROLE
       - PGRST_DB_SCHEMA
+      - PGRST_SERVER_HOST=backend
       - PGRST_SERVER_PORT
+      - PGRST_ADMIN_SERVER_PORT
```
dmijatovic, line 56: Where this env variables is used? I cannot trace it in the source code.

dmijatovic (changes requested):
I think we can make few more health check improvements.

docker-compose.yml
ewan-escience: This is needed to get the [health flag working](https://docs.postgrest.org/en/v14/references/admin_server.html#admin-server).

docker-compose.yml
ewan-escience: I agree that we should do this, but this is out of scope for this PR.

ewan-escience: I will add the `condition: service_healthy` to other services that depend on the backend or database service (but not make new healtchchecks).

dmijatovic (approved):

PR #1822, opened 2026-06-16
Fix drop-and-drag border and unnecessary tab

### LINKED ISSUE #1819: Border behaviour on dragging images
When dragging images (to upload them) on the edit software, edit project and edit news pages, the border appears in a weird two-step animation. This animation should be removed.

Furthermore, in all drag-and-drop places, the appearing border causes a content shift of other parts of the page. We should prevent this (e.g. with an invisible border when no image is being dragged).

### LINKED ISSUE #1820: Unneccessary focus on draggable list elements
List elements that are draggable (e.g. organisations on the edit software page) are focusable with the tab key (and have a border on focus), but they don't have any interactive semantics.

ewan-escience:

* When dragging and dropping an image, the border does not animate any more; neither does it cause a content shift any more
* Remove extra tab that focuses a draggable list element without interactive semantics

### How to test

* Sign in as admin
* Drag and drop images in various places
* Try tab navigating a sortable list, e.g. the organisations of software

closes #1819
closes #1820

PR Checklist:

* [x] Link to a GitHub issue
* [ ] Update documentation
* [ ] Tests

### DIFF
--- a/frontend/components/form/ImageDropZone.tsx
+++ b/frontend/components/form/ImageDropZone.tsx
@@ -1,5 +1,5 @@
+// SPDX-FileCopyrightText: 2025 - 2026 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 // SPDX-FileCopyrightText: 2025 - 2026 Netherlands eScience Center
-// SPDX-FileCopyrightText: 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
 // SPDX-FileCopyrightText: 2026 Dusan Mijatovic (Netherlands eScience Center)
 //
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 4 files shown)


dmijatovic (approved):
Excellent work!

Maybe we could use `var(--rsd-base-700, #232323)` instead of primary color? 
It works well now but it might not work in case the background/page uses primary color.

ewan-escience:
But what if `--rsd-base-700` is used as a background colour? Do we have guidelines on what colour to use when?

dmijatovic:
> But what if `--rsd-base-700` is used as a background colour? Do we have guidelines on what colour to use when?

That is possible too :-). Maybe use `--rsd-base-content` which generally should be always contrasted to background color. In general, primary and secondary colors are used for headers/buttons/background. This gave "character" to the website. These colors also have their contrast partner (rsd-primary-contrast) to ensure contrast color is always garanteed. Then there are content/text colors like rsd-base-content and lastly the rsd-base-xxx colors which are used to add some color variety for borders/lines/frames etc.

I hope this makes sense. Primary color works with the current color scheme and background. Normally I use base-xxx colors for borders but that is more how I prefer to use it.

PR #1826, opened 2026-06-23
feat: split categories filter at top level for organisations

### LINKED ISSUE #1724: Improve the layout of category filter
Currently, the category trees are flattened in the filters. This does not always make sense however. For example, when talking to the ESTRO community, we got questions about this.  

The categories they want to use are:
 
- Field
    - Multiple purpose
    - Photons
    - Protons/Ions
    - Brachyotherapy
- Code type
    - Application
    - Script / Plugin
    - Library / Toolkit

The `Field` one are already in use and look like this:

Image

However, flattening the list of categories to include the top level elements `Field` and `Code Type` does not make sense, as these do not mean anything by themselves. They are just the overall name of the category. 

We should improve this by either: 

- showing the categories as a tree in the filter
- showing a separate filter for each category (using the top level name as a title?)

Just ignoring the top level name `Field` would not work, as this would still mix unrelated categories in one list.

dmijatovic:
# Split categories at the top level for organisations

Closes #1724

* Split organisation categories into separate filters at the top level
* Top level "filters" are ordered on short name (ASCENDING)
* Options in the filter are ordered on short name (ASCENDING)
* Update unit tests with this change
* Note! During categories test it seem that is it NOT possible to remove category items if the category is assigned to software or project. This is intentional, only categories that are not assigned to any software or project can be removed.   

How to test:
* navigate to organisation and confirm that multiple category filters are loaded
* navigate to software page with categories. Click on the category to navigate to organisation software page. Confirm that filter is applied and all filters work properly.
* navigate to project page with categories. Click on the category to navigate to organisation projects page. Confirm that filter is applied and all filters work properly.
* login as rsd admin to be able to view organisation categories. Select organisation and navigate to settings. If the organisation already has categories, confirm that filters on organisation software and project pages show all category filters of the organisation (the top level of category should be the name of the filter).
* if the organisation has no categories create some:
  * When you create top level category without children the filter will not be shown
  * When you create top level category filter, check software and create at lease one child item. The filter should be shown. 
* test other manipulation of organisation categories and how these reflect to custom category filters on organisation page.

## Example organisation projects custom categories
image

PR Checklist:

*   [ ] Update documentation
*   [x] Tests

### DIFF
--- a/database/109-category-functions.sql
+++ b/database/109-category-functions.sql
@@ -1,6 +1,6 @@
--- SPDX-FileCopyrightText: 2024 Dusan Mijatovic (Netherlands eScience Center)
+-- SPDX-FileCopyrightText: 2024 - 2026 Dusan Mijatovic (Netherlands eScience Center)
+-- SPDX-FileCopyrightText: 2024 - 2026 Netherlands eScience Center
 -- SPDX-FileCopyrightText: 2024 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
--- SPDX-FileCopyrightText: 2024 Netherlands eScience Center
 --
 -- SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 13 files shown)


cmeessen (changes requested):
I changed the short names of a category tree that was added by the data generator in an organisation, and then the corresponding category disappeared from the filters.

Image

Image

Update: it appears the filter disappears only when changing the names of leaf categories.

jmaassen:
When sub categories have a name clash (which is likely with large onthologies), the filters will return both, which is not entirely correct. 

For example, we have two categories, each with a subcategory: 

- A
   - X
- B
   - X 

When than have 2 project, each with one category: 

- P1
  - X (of A)
- P2
  - X (of B)  

When I now browse to the page of P1 and click on X, the filter will show me both P1 and P2. It should only show P1?

To get to the expected behavior, I think we would need to filter on the entire category path? Something like "A:X", instead of just "X"

dmijatovic:
@jmaassen This is known/by design approach. High likely we should improve this but we should create new issue for this as it requires substantial effort changes in the database rpcs used for filtering because all filters and search on this page work in concert.

ewan-escience (approved):
The database function looks good and the rest also seems to work. The bug regarding duplicate entries based on filtering on the short name of the category was already present before and should be addressed separately.

dmijatovic:
@cmeessen Can you confirm that categories work for you now?

cmeessen (approved):
The filters now stay when renaming the leafs 👍

PR #1830, opened 2026-06-30
feat: update github actions for authentication and background-services

dmijatovic:
# GitHub Actions CI pipeline for the Java directories

Here is a quick summary, generated by AI, describing changes to help with your code review:

### 1. Split-File Architecture (Fork-Safe Sonar Checks)
* **The Problem:** Standard SonarCloud steps crash instantly on fork PRs because GitHub prevents forks from accessing our repository secrets (`SONAR_TOKEN`).
* **The Solution:** Decoupled our builds into a secure two-step pipeline:
  1. **Primary Workflow (`pull_request`):** Compiles code, executes tests, and uploads the compiled target folder binaries as a safe runner artifact. This step has no privileges and runs flawlessly on forks.
  2. **Sonar Workflow (`workflow_run`):** Triggers automatically on the `main` branch after the primary workflow finishes successfully. Because it executes in our main context, it securely reads `SONAR_TOKEN`, downloads the test artifacts, and publishes metrics to SonarCloud.

### 2. Standardized Reusable Workflows
Instead of copying boilerplate shell scripts across multiple files, everything is abstracted into three centralized workflow engines located under `.github/workflows/`:
* `_pr_info.yml`: Dynamically resolves the active Pull Request context (ID, source branch, and target base branch) via a GitHub REST API query using the unique Commit SHA. This completely fixes edge cases where the fork branch metadata is empty.
* `_java_tests.yml`: Centralizes our standard Java 21 environment setups and executes `mvn clean verify`.
* `_java_sonar.yml`: Orchestrates checking out the detached merge commit securely, mapping the parsed PR inputs, and running the `sonar-maven-plugin`.

### 3. Dependency Caching Optimization
* Removed our standalone `actions/cache@v4` steps for tracking `~/.m2`.
* Migrated to the native language integration feature inside `actions/setup-java@v5` using `cache: 'maven'`. This eliminates boilerplate lines and handles automated path cleanup to prevent build caches from becoming corrupted during parallel module actions.

### 4. Items for You to Verify/Fix (Java Specifics)
I need you to verify two things in our `pom.xml` configs:
* **Missing JaCoCo Coverage:** I noticed `authentication/pom.xml` explicitly defines the `jacoco-maven-plugin` to generate XML reports, but `background-services/pom.xml` does not. Could you add the plugin block to the background services POM? Otherwise, SonarCloud won't be able to calculate or track test coverage metrics for that module.
* **Sonar Project Keys Verification:** I set the keys based on our current workflows, but please double-check them against the SonarCloud Web UI. By default, the Maven plugin builds the key as `groupId:artifactId`. 
  * `authentication` uses `nl.research-software:authentication` (matches its POM).
  * `background-services` currently passes `rsd-background-services`. If the project dashboard in SonarCloud is actually named `nl.esciencecenter:background-services` (matching its POM's groupId), please synchronize it in the `background-services-sonar.yml` or add a `<sonar.projectKey>` property to its POM.

*Note for Review:* Because `workflow_run` events evaluate tasks explicitly from our default branch context, this entire optimization will fully activate on PR branches only after these shared workflow configurations are merged into `main`.
PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/.github/workflows/_java_sonar.yml
+++ b/.github/workflows/_java_sonar.yml
@@ -1,0 +1,99 @@
+# SPDX-FileCopyrightText: 2026 Dusan Mijatovic (Netherlands eScience Center)
+# SPDX-FileCopyrightText: 2026 Netherlands eScience Center
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: reusable java sonar
+
+on:
+  workflow_call:
+    inputs:
+      head_sha:
+        required: true
+        type: string
+      head_branch:
+        required: true
+        type: string
+      working_directory:
+        required: true
+        type: string
+      artifact_name:
+        required: true
+        type: string
+      sonar_project_key:
+        required: true
+        type: string
+      workflow_run_id:
+        required: true
+        type: string
+    secrets:
+      SONAR_TOKEN:
+        required: true
+
+jobs:
+  # 1. Fetch PR info in order to determine where Sonar check should run
+  resolve-pr:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    uses: ./.github/workflows/_pr_info.yml
+    with:
+      head_sha: ${{ inputs.head_sha }}
+
+  # 2. Run Sonar analysis using inputs and reslove-pr step
+  scan:
+    name: Java Sonar Scanner
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-24.04
+    needs: resolve-pr
+    steps:
+      - name: Checkout Code Securely
+        uses: actions/checkout@v6
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
+          ref: ${{ inputs.head_sha }}
+          fetch-depth: 0
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+          cache: 'maven'
+
+      - name: Download Target Artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ inputs.artifact_name }}
+          path: ${{ inputs.working_directory }}/target/
+          run-id: ${{ inputs.workflow_run_id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+
+      - name: Run Sonar Analysis (PR Mode)
+        if: ${{ needs.resolve-pr.outputs.is_pr == 'true' }}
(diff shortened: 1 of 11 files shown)


.github/workflows/_java_sonar.yml
```diff
@@ -0,0 +1,98 @@
...
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    uses: ./.github/workflows/_pr_info.yml
+    with:
+      head_sha: ${{ inputs.head_sha }}
+
+  # 2. Run Sonar analysis using inputs and reslove-pr step
+  scan:
+    name: Java Sonar Scanner
+    runs-on: ubuntu-24.04
+    needs: resolve-pr
```
ewan-escience, line 46: According to [the documentation on `if`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions):

> A job that is skipped will report its status as "Success".

And according to [the documentation on `needs`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds):

> Use jobs.<job_id>.needs to identify any jobs that must complete successfully before this job will run.

So it seems to me that step 2 will also run if step 1 is skipped, which is not what you want I think.

.github/workflows/_java_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Execute Tests and Generate Reports
+        working-directory: ${{ inputs.working_directory }}
+        run: mvn -B clean verify
+
+      - name: Upload Target Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact_name }}
+          path: ${{ inputs.working_directory }}/target/
+          retention-days: 1
```
ewan-escience, line 46: It is unclear to me if the name needs to be unique only per workflow or per repo. In the latter, two simultaneously opened PRs could collide. [This issue[(https://github.com/actions/upload-artifact/issues/668) doesn't provide the answer... I'd like to see a source stating that this is ok.

.github/workflows/authentication_tests.yml
```diff
@@ -20,34 +21,8 @@ on:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        working-directory: authentication
-        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=nl.research-software:authentication -Pcoverage
+    name: authentication tests
+    uses: ./.github/workflows/_java_tests.yml
+    with:
+      working_directory: "authentication"
+      artifact_name: "authentication-target-dir"
```
ewan-escience: Please make sure to add a newline to the end of files.

ewan-escience (changes requested):
I think there is one logic error. For the rest, it looks like a good approach.

ewan-escience (approved):
Looks good, let's give it a go!

PR #1831, opened 2026-06-30
Add support for software badges

### LINKED ISSUE #1740: Link to EVERSE software quality dashboard (EVERSE Milestone 11)
As part of the [EVERSE project](https://everse.software), we have promised to implement the following milestone (due at 31/08/2026): 

*Demonstration of the outcome of the common WP3 framework to the research community, by showing software quality indicators in existing services (such as OpenEBench, WorkflowHub, Research Software Directory, ENVRIHUB, etc) based on the dashboard prototype.*

In EVERSE, @fdiblen has developed a software quality dashboard, which can be found at: https://github.com/EVERSE-ResearchSoftware/DashVERSE 

In practice this milestone comes down to the following for the RSD: 

- we would need to be able to add one or more 'software quality badges' to a software page
- this badge shows up on the software page somewhere (probably the sidebar)
- if the badge is clicked, a tab will open taking you the service that shows the detailed information

This is very similar to the badges we also use in the README.md of the RSD itself, for example: 

[CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/6336)

We can discuss if we will just allow the addition of arbitrary badges (as shown above), of if we would want to add well known URLs that allow us to harvest information (like we do with github, pypi, etc).

ewan-escience:

* Add support for badges on software pages
* Additional validation added for verification if badge URL returns an (valid) image.
* Added basic unit tests

### How to test

* `docker compose --profile "*" down --volumes && docker compose --profile data build --parallel && docker compose --profile data up`
* Sign in as admin
* Add, edit and reorder badges of software pages (badges are also created with the data-generation script)
* Check how badges are rendered on the software pages
* Check if you agree with the design of the database

closes #1740

PR Checklist:

* [x] Link to a GitHub issue
* [x] Update documentation
* [x] Tests


jmaassen (approved):
From a user perspective this works as expected.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+export default function EditSoftwareBadges() {
+  const {token} = useSession()
+  const {showErrorMessage} = useSnackbar()
+  const {setValue, watch} = useFormContext<EditSoftwareItem>()
+
+  const [isAddBadgeModalOpen, setIsAddBadgeModalOpen] = useState<boolean>(false)
+  const [badgeToEdit, setBadgeToEdit] = useState<BadgeForSoftware | null>(null)
+
+  const softwareId = watch('id')
+  const badges: BadgeForSoftware[] = watch('badges')
```
dmijatovic: (Recomended) This can be written in one line, it seem to me one function call less.

```typescript
const [softwareId, badges] = watch(['id','badges'])
```

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+import EditSoftwareBadgeModal from '~/components/software/edit/links/EditSoftwareBadgeModal'
+import Tooltip from '@mui/material/Tooltip'
+
+export default function EditSoftwareBadges() {
+  const {token} = useSession()
+  const {showErrorMessage} = useSnackbar()
+  const {setValue, watch} = useFormContext<EditSoftwareItem>()
+
+  const [isAddBadgeModalOpen, setIsAddBadgeModalOpen] = useState<boolean>(false)
+  const [badgeToEdit, setBadgeToEdit] = useState<BadgeForSoftware | null>(null)
```
dmijatovic: (Optional) I would combine this two states into one state for modal as these always go together, although for new item the data is optional.

```typesctipt
const [modal, setModal] = useState<{
    open: boolean,
    data?:BadgeForSoftware
  }>({
    open: false
  })
```

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const newBadgesArray = badges.toSpliced(deletedBadgeIndex, 1)
+
+        for (let i = 0; i < newBadgesArray.length; i++) {
+          newBadgesArray[i].position = i + 1
+        }
+
+        handleSortedBadges(newBadgesArray)
+      })
+      .catch(e => showErrorMessage(e.message))
+  }
```
dmijatovic: (Recommended) I prefer to have visual/presentation (html part) and working/api part (functions) to be separated (as much as possible). In modern React this means that functions are moved to "hook" file, imported and passed to child components if needed. Usually the state will be also in the hook file.

In addition,  `allSettled` does not trow exception, the catch will not work. The proper approach with allSettled looks something like this

```Typescript
 Promise.allSettled(promises).then((results) => {
    const errors = results.filter(r => r.status === 'rejected')
    if (errors.length > 0) showErrorMessage('Failed to reorder items.')
  })
```

In handle delete the most efficient way to remove item is using filter asfik.
```typescript
const newBadgesArray = badges.filter(badge => badge.id !== badgeId))
```
I do not think that updating all position is needed. We order by position but the values do not need to be sequential. And by the next reorder the position will be updated to sequential numbers. The danger of reordering with delete is that error might confuse user if it is not in delete action but during updating positions (I think the action is not transparent to user).

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  function renderBadge(badge: BadgeForSoftware) {
+    const badgeContent = <img src={badge.badge_url} alt={badge.badge_url} className="max-h-[20px]" />
```
dmijatovic: (Required) Please add descriptive text to alt like `alt="Software badge X"` or just an empty string to prevent screen reader from reading a link to user. We recently removed all url from alt based on a11y audit.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    const secondaryAction = (
+      <>
+        {badge.link_url && <Tooltip placement="top" title={`Go to ${badge.link_url}`}><IconButton LinkComponent={Link} href={badge.link_url}><LinkIcon /></IconButton></Tooltip>}
+        <Tooltip placement="top" title="Edit badge"><IconButton onClick={() => {openEditBadgeModal(badge)}}><EditIcon /></IconButton></Tooltip>
+        <Tooltip placement="top" title="Delete badge"><IconButton onClick={() => handleDeleteBadge(badge.id)}><DeleteIcon /></IconButton></Tooltip>
```
dmijatovic: (Recommended) I would make this simpler. For the button label I would prefer default Save button we use in other modals.

In addition, the code is difficult to read for a reviewer. I would appreciate if you write one component per line and indent child component. 

Now -> difficult to read/review

```Typescript
{badge.link_url && <Tooltip placement="top" title={`Go to ${badge.link_url}`}><IconButton LinkComponent={Link} href={badge.link_url}><LinkIcon /></IconButton></Tooltip>}
        <Tooltip placement="top" title="Edit badge"><IconButton onClick={() => {openEditBadgeModal(badge)}}><EditIcon /></IconButton></Tooltip>
        <Tooltip placement="top" title="Delete badge"><IconButton onClick={() => handleDeleteBadge(badge.id)}><DeleteIcon /></IconButton></Tooltip>
```

preffered -> easier to review and understand what is happening

```Typescript
<>
  {badge.link_url &&
    <Tooltip placement="top" title={`Go to ${badge.link_url}`}>
      <IconButton LinkComponent={Link} href={badge.link_url}>
        <LinkIcon />
      </IconButton>
    </Tooltip>
  }
  <Tooltip placement="top" title="Edit badge">
    <IconButton onClick={() => {openEditBadgeModal(badge)}}>
      <EditIcon />
    </IconButton>
  </Tooltip>
  <Tooltip placement="top" title="Delete badge">
    <IconButton onClick={() => handleDeleteBadge(badge.id)}>
      <DeleteIcon />
    </IconButton>
  </Tooltip>
</>
```

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+    <>
+      <EditSectionTitle
+        title={config.badges.title}
+        subtitle={config.badges.subtitle}
+      >
+        <Button
+          variant='contained'
+          onClick={() => openEditBadgeModal(null)}>
+          Add badge
+        </Button>
```
dmijatovic: (Required) Please add + icon. See example of Add button in Source code section of edit software.

Image

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+  deleteBadge,
+  getBadgesForSoftware,
+  updateBadgePosition
+} from '~/components/software/apiSoftware'
+import IconButton from '@mui/material/IconButton'
+import ListItemText from '@mui/material/ListItemText'
+import Link from 'next/link'
+import {useState} from 'react'
+import EditSoftwareBadgeModal from '~/components/software/edit/links/EditSoftwareBadgeModal'
+import Tooltip from '@mui/material/Tooltip'
```
dmijatovic: (Optional) I usually organize imports in the following order:
- react
- next
- material
- other external libraries
- internal not in the same folder/section
- internal in the same folder/section

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+          Add badge
+        </Button>
+      </EditSectionTitle>
+      {isAddBadgeModalOpen && <EditSoftwareBadgeModal
+        softwareId={softwareId}
+        existingBadgeUrls={urlsPresent}
+        onSave={() => handleNewBadgeAdded()}
+        onCancel={() => setIsAddBadgeModalOpen(false)}
+        badgeToEdit={badgeToEdit === null ? null : {badgeId: badgeToEdit.id, badgeUrl: badgeToEdit.badge_url, badgeLink: badgeToEdit.link_url}}
+      />}
```
dmijatovic: (Recommended) Use standard ternary approach `isAddBadgeModalOpen ? <component>:null`. 

"The && operator in JavaScript evaluates to the first falsy value it encounters, or the last value if all are truthy.If your condition evaluates to a falsy boolean (false), React safely renders nothing. However, if your condition evaluates to a falsy number (0), JavaScript returns 0. React then tries to render that number, causing a rogue 0 to appear directly on your screen."

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <SortableList
+        items={badges}
+        onSorted={(sortedBadges) => handleSortedBadges(sortedBadges)}
```
dmijatovic: This function can be directly passed to component as the params are the same (one function call less)

```typescript
...
onSorted={handleSortedBadges}
...

```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+import useSnackbar from '~/components/snackbar/useSnackbar'
+
+
+export type EditSoftwareBadgeModalProps = {
+  softwareId: string,
+  existingBadgeUrls: Set<string>,
+  onSave: () => void,
+  onCancel: () => void,
+  badgeToEdit: EditBadgeFields | null,
+}
```
dmijatovic: (Optional) You can add `Readonly` as part of initial type. For me the function signature reads easier (its shorter). In the case of Modals that support new and edit I usually have data as optional argument not as null. What advantages you see using null instead of optional argument (undefined)?

```typescript
export type EditSoftwareBadgeModalProps = Readonly<{
  softwareId: string,
  existingBadgeUrls: Set<string>,
  onSave: () => void,
  onCancel: () => void,
  badgeToEdit?: EditBadgeFields,
}>
```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+    } else {
+      return enteringNewBadge ? 'Add badge' : 'Update badge'
+    }
+  }
+
+  return (
+    <Dialog open={true}>
+      <DialogTitle>
+        Add new badge
+      </DialogTitle>
```
dmijatovic, line 107: (Required) Please use title formatting as with other modals ) `sx`, like in SoftwareRepositoryModal

```typescript
<DialogTitle sx={{
  fontSize: '1.5rem',
  borderBottom: '1px solid',
  borderColor: 'divider',
  color: 'primary.main',
  fontWeight: 500
}}>
    {title}
</DialogTitle>
```

The style remarks concerning this component should result in the "look and feel" as used in the SoftwareRepositoryModal component.

Image

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <form onSubmit={handleSubmit(handleUpdateOrAddBadge)} autoComplete='off'>
+        <DialogContent>
```
dmijatovic: (Required) Please use styles as in SoftwareRepositoryModal.

```typescript
...
<DialogContent sx={{
    width: ['100%', '37rem'],
    padding: '2.5rem 1.5rem',
    display: 'flex',
    flexDirection: 'column',
    gap: '1.5rem'
  }}>
...
```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+  function renderAddButtonText(): string {
+    if (isSubmitting) {
+      return 'Saving'
+    } else {
+      return enteringNewBadge ? 'Add badge' : 'Update badge'
+    }
+  }
+
+  return (
+    <Dialog open={true}>
```
dmijatovic: (Required) Please use dialog settings for mobile and the useSmallScreen hook. See <SoftwareRepositoryModal> component and props passed to <Dialog ...> there.

On  mobile screen the dialog should open full screen, now it looks like in the pic below
Image

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <DialogTitle>
+        Add new badge
+      </DialogTitle>
+      <form onSubmit={handleSubmit(handleUpdateOrAddBadge)} autoComplete='off'>
```
dmijatovic: (Required) Please assign unique id to form. The form submit works now but if we need to pull submit button outside of the form in the future submit button will not work. It is recommended to use id on every form.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+            autoFocus={true}
+            {...register('badgeUrl', {...config.badges.validation.badgeUrl, validate: validateNewUrlNotExistsYet})}
+            defaultValue={enteringNewBadge ? '' : badgeToEdit!.badgeUrl}
+            fullWidth
+            margin='normal'
+            label={'URL of the image of the badge'}
+            error={!!errors.badgeUrl}
+            helperText={errors.badgeUrl?.message}
+            required={true}
+          >
```
dmijatovic: (Required) Please use generic component `ControlledTextField` we created for text fields that should show max size of the file and size of current string. If you need more information you can have a look at SoftwareRepositoryModal.tsx. 

In addition
- please show initial help info in order to avoid layout shift on error.
- please add all labels, info and validation to config as much as possible. Use of variables instead of plain text is prefered.

Here is example of ContolledTextField for badgeUrl.

```typescript
<ControlledTextField
  control={control}
  options={{
    name: 'badgeUrl',
    autofocus: true,
    label: config.badges.badgeUrl.label,
    useNull: true,
    defaultValue: badgeToEdit?.badgeUrl ?? null,
    helperTextMessage: errors['badgeUrl']?.message ?? config.badges.badgeUrl.help,
    helperTextCnt: `${getValues('badgeUrl')?.length ?? 0}/${config.badges.badgeUrl.validation.maxLength.value}`,
    // endAdornment: loading ?
    //   <CircularProgress data-testid="slug-circular-progress" color="primary" size={32} />
    //   : undefined
  }}
  rules={{
    ...config.badges.badgeUrl.validation,
    validate: validateNewUrlNotExistsYet
  }}
/>

```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <TextField
+            {...register('badgeLink', {...config.badges.validation.badgeLink})}
+            defaultValue={enteringNewBadge ? null : badgeToEdit!.badgeLink}
+            fullWidth
+            margin='normal'
+            label={'Optional link of the badge'}
+            error={!!errors.badgeLink}
+            helperText={errors.badgeLink?.message}
+            required={false}
+          >
```
dmijatovic: Same as previous input.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+          </TextField>
+        </DialogContent>
+        <DialogActions>
```
dmijatovic: (Required) Please add styles from other modal

```typescript
 <DialogActions sx={{
    padding: '1rem 1.5rem',
    borderTop: '1px solid',
    borderColor: 'divider'
  }}>
```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,132 @@
...
+            disabled={!isValid || isSubmitting || isSubmitted}
+            type='submit'
+          >
+            {renderAddButtonText()}
+          </Button>
+          <Button
+            variant='outlined'
+            onClick={() => onCancel()}>
+            Cancel
+          </Button>
```
dmijatovic, line 172: (Required) Please use same buttons as SoftwareRepositoryModal. The primary button, submit, in this case is always most right button. Secondary button does not have border. The tab order is: primary , secondary button. I expect that user used SoftwareRepositoryModal before coming here and would expect similar behaviour using this modal.

dmijatovic (changes requested):
First, you have done lot of work in this PR! I appreciate your contribution but I would like to have more "standardized" approach used in other modals and in the rest of the frontend codebase. 

I have classified my remarks in optional, recommended and required. I think the most efficient way to approach my remarks is for you to have a look at my remarks and then have a call to discuss these verbally.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: I'm not sure about this one. If a link as alt text is undesirable, maybe we should let the user supply a mandatory alt text.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: What do you mean? Not to use `Tooltip`?

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: I think we need a tool then to automatically sort imports.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: But `isAddBadgeModalOpen` is a `boolean`.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: > What advantages you see using null instead of optional argument (undefined)?

I want the caller of the function to consciously make a choice about whether such a field is present or empty.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: Why are we adding the same style everywhere to the same component? Wouldn't [styleOverrides](https://mui.com/material-ui/customization/theme-components/#theme-style-overrides) make this a single source of truth?

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: > The primary button, submit, in this case is always most right button.

This results in a non-logical tab order, which even results in buggy tab behaviour (on both Firefox and Chromium). Try it out with tabbing and reverse-tabbing.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: I'm not principally against this, but when you need to move the submit to outside the form, you need to add the [`form` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/form) to the button anyway, so you won't forget to add an ID to the form.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: Yes, that could be an option. I assume that option might be applicable in few more places.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: Actually I was confused with interactive change of the button label for Save. This code was not that readable. After changing the layout I could better understand what is happening :-). I prefer default title property because I think the browser decides the positioning of the text on mouse over while using Tooltip component requires "fixed" position of the text.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: Well spotted! So we will need to fix this in all modals :-). 

```typescript
<DialogActions sx={{
  padding: '1rem 1.5rem',
  borderTop: '1px solid',
  borderColor: 'divider',
  // Flip button position
  flexDirection: 'row-reverse',
  // Start from right to left
  justifyContent: 'flex-start',
  gap:'1rem'
}}>
  <SubmitButtonWithListener
    formId={formId}
    disabled={!isValid || isSubmitting || isSubmitted}
  />
  <Button
    color="secondary"
    onClick={onCancel}>
    Cancel
  </Button>
</DialogActions>
```

**Maybe we should create shared components for dialog with all these styles appended, something like RsdDialogTitle, RsdDialogContent and RsdDialogActions and then use these in all modals.**

dmijatovic: Can the disabled validation `!isValid || isSubmitting || isSubmitted` extend to include isDirty? For example if you start edit with "Value 1" then you change to "Value 2" isDirty=true, but if you revert back to "Value 1" isDirty=false. When isDirty is false Save should be disabled = there is nothing new to save.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: I am not sure if the "tool" can perform desired order import by library (and then using specific hierarchy generic-> more specific -> local by folder "distance" from the file), usually they order alphabetically asfik. I find it not worth extra dependency to manage/maintain.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: There could be situations that 0 is rendered instead of nothing

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: I agree for "field" with simple properties, these could be value or null. This approach works perfectly with backend and databases. The objects should be defined/undefined in my opinion and the specific checks and default values should be assigned when needed using pattern `object?.property ?? default value`.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: We decided to let the user enter an optional alt text.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: Yes, that could be also good approach. We can try that or create custom components!

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: I agree, but I would like you to use standard submit button which requires formId ;-), although we can change that param to be optional :-).

ewan-escience: What standard submit button do you mean?

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: For now I will copy over the style (and media queries as you described in a different comment), but we should make an issue in which we describe the need for shared styling or custom components.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: SubmitButtonWithListener is used in other modals, for example SoftwareRepositoryModal

```typesctipt
<SubmitButtonWithListener
  formId={formId}
  disabled={isSaveDisabled()}
/>
```

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: > I do not think that updating all position is needed. We order by position but the values do not need to be sequential. And by the next reorder the position will be updated to sequential numbers. The danger of reordering with delete is that error might confuse user if it is not in delete action but during updating positions (I think the action is not transparent to user).

But then adding an new badge might not append it to the bottom, confusing the user even more. If an error happens during updating the positions, I would like the user to know something went wrong and hopefully report it to us.

dmijatovic: You have a valid point there, although at other places this is not implemented we can implement it here. You can "chain" map on filter and call patch.

```Typescript
//remove item and renumber positions
const newBadgesArray = badges.filter(badge => badge.id !== badgeId))
           .map((item, pos) => ({ ...item, position: pos + 1 }))
//patch all positions
handleSortedBadges(newBadgesArray)
```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const {showErrorMessage} = useSnackbar()
+  const {watch, getValues, handleSubmit, control, formState: {errors, isValid, isDirty, isSubmitting, isSubmitted}} = useForm<NewBadgeFields>({
+    mode: 'onChange',
+    defaultValues: {
+      badgeUrl: badgeToEdit?.badgeUrl ?? '',
+      altText: badgeToEdit?.altText ?? '',
+      badgeLink: badgeToEdit?.badgeLink ?? '',
+    }
+  })
+  const enteringNewBadge: boolean = badgeToEdit === null
```
dmijatovic: Optional: maybe just use `badgeToEdit === null` in the conditionals further in the form

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+import SubmitButtonWithListener from '~/components/form/SubmitButtonWithListener'
+import ControlledTextField from '~/components/form/ControlledTextField'
+
+
+export type EditSoftwareBadgeModalProps = {
+  softwareId: string,
+  existingBadgeUrls: Set<string>,
+  onSave: () => void,
+  onCancel: () => void,
+  badgeToEdit: EditBadgeFields | null,
```
dmijatovic: For me `badgeToEdit` is optional property, it exists on edit and not on new item. I would strongly advice to use "closer to reality" type 

```Typescript
export type EditSoftwareBadgeModalProps = {
  softwareId: string,
  existingBadgeUrls: Set<string>,
  onSave: () => void,
  onCancel: () => void,
  badgeToEdit?: EditBadgeFields,
}
```

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (enteringNewBadge) {
+      return badgeUrl === null || !existingBadgeUrls.has(badgeUrl) ? true : 'This badge already exists for this software'
+    }
+    else {
+      return badgeUrl === null || badgeUrl === badgeToEdit!.badgeUrl || !existingBadgeUrls.has(badgeUrl) ? true : 'This badge already exists for this software'
+    }
+
+  }
+
+  function handleUpdateOrAddBadge() {
```
dmijatovic: **Required**: Please use data directly passed by `handleSubmit` in the function you are giving to handleSubmit.

**Recommended**: Why `updateBadgeContent` and `addBadgeForSoftware` use different param names for same data values? I would prefer to use same names when same props are passed. The name "conversion", if needed, should take place in api module when getting or saving data. Ideally, "properties" shold be consistently converted as much as possible in following way db: snake_case -> fe: camelCase; example:  db_name -> dbName.

**Optional**: I would use different decision logic for add/edit based on `badgeToEdit` id presence. Then you do not need to `force!` the type and you can easily provide default value. 

**Optional**: I advice latest/modern async/await pattern with try/catch/finally instead of original, but still working/valid, Promise approach.

**Optional**: the function name could also be "handleOnSubmit"

```Typescript
async function handleUpdateOrAddBadge(data:NewBadgeFields) {
    try{
      if (badgeToEdit?.badgeId){
        await updateBadgeContent(token, {
          badgeId: badgeToEdit.badgeId,
          badgeUrl: data.badgeUrl,
          badgeLink: data.badgeLink,
          altText: data.altText,
        })
      }else {
        await addBadgeForSoftware(token, {
          software: softwareId,
          badge_url: data.badgeUrl,
          link_url: data.badgeLink,
          position: existingBadgeUrls.size + 1,
          alt_text: data.altText,
        })
      }
    }catch(e:any){
      showErrorMessage((e.message))
    }finally{
      onSave()
    }  
}
````

File: frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx [hunk 22]
```diff
@@ -0,0 +1,168 @@
...
+          padding: '1rem 1.5rem',
+          borderTop: '1px solid',
+          borderColor: 'divider',
+          flexDirection: 'row-reverse',
+          justifyContent: 'flex-start',
+          gap:'1rem',
+        }}>
+          <SubmitButtonWithListener
+            formId={formId}
+            disabled={!isValid || isSubmitting || isSubmitted || !isDirty}
```
dmijatovic: Optional: `!isValid || !isDirty` is sufficient?

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+          />
+          <Button
+            variant='outlined'
```
dmijatovic: Required: Please remove `variant='outlined'` from Cancel button to be in line with other modals.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,140 @@
...
+          newBadgesArray[i].position = i + 1
+        }
+
+        handleSortedBadges(newBadgesArray)
+      })
+      .catch(e => showErrorMessage(e.message))
+  }
+
+  function renderBadge(badge: BadgeForSoftware) {
+    const badgeContent = <img src={badge.badge_url} alt={badge.alt_text ?? ''} className="max-h-[20px]" />
```
dmijatovic, line 38: Recommended: Please open link in new tab and use `rem` instead of `px`. The design scales/steps/sizes are usually 4/8/12/16/20/24/32px where 16px = 1rem. Most often used sizes are 16px = 1rem, 24px = 1.5rem, 32px = 2rem. The symmetry and "empty" space is an important part of "good looking" design.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
```diff
@@ -0,0 +1,140 @@
...
+        const newBadgesArray = badges.filter(badge => badge.id !== badgeId)
+
+        for (let i = 0; i < newBadgesArray.length; i++) {
+          newBadgesArray[i].position = i + 1
+        }
+
+        handleSortedBadges(newBadgesArray)
+      })
+      .catch(e => showErrorMessage(e.message))
+  }
```
dmijatovic: Recommended: This complete section of functions can be moved to hook `useEditSoftwareBadges` and only imported and used here. Most react developers find the hook approach convinient and in the rest of the code we mainly use hook approach. [See this branch for example implementation.](https://github.com/research-software-directory/RSD-as-a-service/blob/1740-software-badges-dm/frontend/components/software/edit/links/useEditSoftwareBadges.tsx)

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+        }}>
+          <ControlledTextField
+            control={control}
+            rules={{...config.badges.badgeUrl.validation, validate: validateNewUrlNotExistsYet}}
+            options={{
+              name: 'badgeUrl',
+              label: config.badges.badgeUrl.label,
+              autofocus: true,
+              helperTextMessage: errors['badgeUrl']?.message ?? config.badges.badgeUrl.help,
+              helperTextCnt: `${watch('badgeUrl')?.length ?? 0}/${config.badges.badgeUrl.validation.maxLength.value}`
```
dmijatovic: Optional: maybe combine into one watch for all three values at the top of file? There is also `useWatch` hook that might fit better for this usecase, but I am not sure if it makes any difference in this particular case.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: Optional: The modern react version of this function looks usually like example below. `Filter` is usually used in combination with `map` or `reduce` to perform any additional changes on remaining items and **return new array**. `For loop` manipulates values within existing array object (which is already registered in useState). The moment you remove filter, which returns new array, your setState will not react (trigger render/refresh) because the same array object reference is passed to setState (hence react will think that the values are the same and there is no reason to re-render component).

```Typescript
async function handleDeleteBadge(badgeId: string){
  try {
    await deleteBadge(token, badgeId);
    
    const updatedBadges = badges
      .filter(badge => badge.id !== badgeId)
      .map((badge, index) => ({
        ...badge,
        position: index + 1
      }));

    handleSortedBadges(updatedBadges);
  } catch (err:any) {
    const message = err?.message : "An unknown error occurred";
    showErrorMessage(message);
  }
}
```

dmijatovic (changes requested):
Nice work! We are making progress step by step. This feature is packed with demanding code and is excellent for further developing of you Frontend React skills. 

There are few points for you to reach "senior" level:
- Use of hooks, separating UI/Presentation component and the functions doing work
- Use of functional array manipulation with filter/map/reduce and object spread. All mentioned methods return new object which is crucial for proper triggering of reactive state (useState hook).
- Using modern async/await function calls instead of original Promise approach.

At the later stage we can extend your knowledge with writing the unit/integration tests :-)

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: But this conveys the meaning. Having to read and interpret `badgeToEdit === null` all the time, will take more effort I presume.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
ewan-escience: No, the save button should be disabled after clicking to prevent duplicate submits (especially when adding a new badge).

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: > use `rem` instead of `px`.

Understandable, but the badges I encountered all have a height of 20px, independently of how we define our `rem`.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: It could be. However my general advice is not to type object as null. In your further decisions use badgetToEdit?.id (has id or not). So actually this comment is rather obsolete when other remarks are taken into a count. Sorry for the confusion.

frontend/components/software/edit/links/EditSoftwareBadgeModal.tsx
dmijatovic: OK. I was not able to trigger isSubmitting || isSubmitted states without form disapearing but theoretically these states could be present in a slow connection situation.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: But the `filter` object already returns a new array, which is used to update the state later on in `handleSortedBadges`. My point is to _prevent_ yet another array creation with `map` and to prevent unnecessary object copying with the spread operator. So I don't think your argument holds. Or are you against mutation of the badge objects?

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: In that case,  I would use `1.25rem` but 20px is also fine :-)

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: I prefer not to use mutation but I understand that sometimes it might be needed. Looking at the code one sees less used approach. I would write it as stated in example above but this is one of the cases where both approaches are valid and you are aware that filter already returns new object. So adding a comment above for loop like "New array is already created using for loop for maximum performance" might be right way to go and also clear for less experienced react programmers :-).

dmijatovic: Actually, if you plan to use for loop and have max performance than filter work can be also performed there :-). 

In "functional world" achieving same result with one loop can be done with `reduce`:
```Typescript
const updatedBadges = badges.reduce((acc, badge) => {
  if (badge.id !== badgeId) {
    acc.push({
      ...badge,
     // use new array's length as the index
      position: acc.length + 1 
    });
  }
  return acc;
}, []);
```

frontend/components/software/edit/links/EditSoftwareBadges.tsx
ewan-escience: But if the `rem` changes, the value here should _not_ change.

dmijatovic: In the past I saw rem approaches used to scale items between desktop/tablet/mobile. We do not use this functionality but in general rem is advised because it offers extra functionality. Anyway I am sure we used px in some other places so this use of rem is not that strict. I provide you a reason why I, and most of CSS designers, are using rem instead of pixels in modern design.

ewan-escience: Because, like I said before, the badges have an intrinsic height of 20px.

dmijatovic: If the badges are svg (scalable-vector-graphic) then I think these will scale properly. But let's leave this as it is.

frontend/components/software/edit/links/EditSoftwareBadges.tsx
dmijatovic: We used for loop for max speed and easy readability for non-functional programmers :-).

dmijatovic (approved):
@ewan-escience Thank you very much for your contribution! 
I have moved "actions" to hook and made additional validation about badge url returning an image as you originally proposed. I think badge image validation is useful feature.

@jmaassen Maybe double check the final version of the feature if you have time?

dmijatovic:
@ewan-escience, @jmaassen This PR is ready for final check.

jmaassen:
Screenshot_2026-07-17_11-04-38

The titles "Badge URL" and "Badge link" are a bit confusing. How about replacing these with "Badge image URL" and "Badge target link" ? 

Also, the alt text box is now the only one without a title?

dmijatovic:
@ewan-escience, @jmaassen Labels are updated and badge image validation is improved to make only one check. Please double check and confirm.

jmaassen (approved):
Works as expected from a user perspective!

ewan-escience:
Thanks for the improvements, I think we can merge this now.

PR #1835, opened 2026-07-03
Update docker-compose.yml for development

### LINKED ISSUE #1689: Improve handling of containers by using profiles
For new developers, the current configuration of docker compose can be a bit confusing, e.g. as it is unclear which containers are required for running the RSD in certain configurations. The usage of `--scale` also implies that some containers might be deployable several times, which is not the case.

Instead, using [profiles](https://docs.docker.com/compose/how-tos/profiles/) could be a viable alternative. Containers are assigned specific profiles which can be addressed when starting up the RSD:

```
docker compose --profile debug up
```

dmijatovic:
# Update docker compose for optional services

Closes #1689

Newer versions (June 2026) of Docker Compose (v5.3.0 and above, including Docker Engine v29+) strictly validate the entire service dependency tree before starting. This validation caused my local development environment to crash with `mail is missing dependency rabbitmq` because those services were set to `replicas: 0`.

* Created Dedicated Profiles:
  * **mail**: `rabbitmq`, `mail`, and `publisher` are now under the mail profile.
  * **data**: `data-generation` is now under its own data profile.
  * **scrapers**: `scrapers` is under its own scrapers profile.
* Added `profiles` tags to completely hide these blocks from modern Docker validation checks unless explicitly requested.
* Removed deploy - replicas from optional services
* Updated Makefile 
    * `make clean`: to clean all profiles and remove volumes.
    * `make start`: cleans everything, starts basic services and generates test data.
    * `make scrapers`: cleans everything, starts basic services and scrapers **but** no test data.

How to test:
1. Run `docker compose up` — verify that only 9 core services start up cleanly without any dependency errors.
2. Verify run of the data generation command manually:
 ```bash
 docker compose --profile data up --detach
 ```
3. Verify you can run mail services with the rest too
```bash
docker compose --profile mail up --detach
```
4. Verify you can run scrapers with the rest
```bash
docker compose --profile scrapers up --detach
```
5. Validate Makefile commands 
   * `make start`  command runs all basic services and generates test data while mail services are not called
   * `make scrapers` command runs basic service and scrapers service
   * `make mail` command runs basic services and mail profile services

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,9 @@
 # SPDX-FileCopyrightText: 2022 - 2024 Christian Meeßen (GFZ) <christian.meessen@gfz-potsdam.de>
 # SPDX-FileCopyrightText: 2022 - 2024 Helmholtz Centre Potsdam - GFZ German Research Centre for Geosciences
 # SPDX-FileCopyrightText: 2022 - 2025 Ewan Cahen (Netherlands eScience Center) <e.cahen@esciencecenter.nl>
-# SPDX-FileCopyrightText: 2022 - 2025 Netherlands eScience Center
+# SPDX-FileCopyrightText: 2022 - 2026 Netherlands eScience Center
 # SPDX-FileCopyrightText: 2022 Jesús García Gonzalez (Netherlands eScience Center) <j.g.gonzalez@esciencecenter.nl>
-# SPDX-FileCopyrightText: 2023 Dusan Mijatovic (Netherlands eScience Center)
+# SPDX-FileCopyrightText: 2023 - 2026 Dusan Mijatovic (Netherlands eScience Center)
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -30,7 +30,7 @@ export DGID
 # ----------------------------------------------------------------
 start: clean
 	docker compose build # build all services
-	docker compose up --scale data-generation=1 --scale scrapers=0 --detach
+	docker compose --profile data up --scale data-generation=1 --detach
 	# open http://localhost to see the application running
 
 install: clean
@@ -44,7 +44,7 @@ install: clean
 	# All dependencies are installed. The data generation module is running in the background. You can now run `make dev' to start the application
 
 clean:
-	docker compose down --volumes
+	docker compose --profile mail --profile data down --volumes
 
 stop:
 	docker compose down
--- a/Makefile
+++ b/Makefile
@@ -29,34 +29,37 @@ export DGID
 # Main commands
 # ----------------------------------------------------------------
 start: clean
-	docker compose build # build all services
-	docker compose --profile data up --scale data-generation=1 --detach
+	docker compose --profile data build
+	docker compose --profile data up --detach
+	# open http://localhost to see the application running
+
+scrapers:	clean
+	docker compose --profile scrapers build
+	docker compose --profile scrapers up --detach
+	# open http://localhost to see the application running
+
+mail: clean
+	docker compose --profile mail build
+	docker compose --profile mail up --detach
 	# open http://localhost to see the application running
 
 install: clean
-	docker compose build database backend auth codemeta scrapers nginx   # exclude frontend and wait for the build to finish
-	docker compose up --scale scrapers=0 --detach
+	# exclude frontend and wait for the build to finish
+	docker compose build database backend auth codemeta scrapers nginx
 	cd frontend && npm install
 	cd documentation && npm install
-	# Sleep 10 seconds to be sure that docker compose up is running
-	sleep 10
-	docker compose up --scale data-generation=1 --detach
-	# All dependencies are installed. The data generation module is running in the background. You can now run `make dev' to start the application
+	docker compose --profile data up --detach
+	# all dependencies are installed. The data generation module is running in the background. You can now run `make dev' to start the application
 
 clean:
-	docker compose --profile mail --profile data down --volumes
+	docker compose --profile "*" down --volumes
 
 stop:
-	docker compose down
+	docker compose --profile "*" down
 
 frontend-dev: frontend/.env.local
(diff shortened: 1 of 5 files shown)


ewan-escience:
See https://github.com/docker/compose/issues/13899.

ewan-escience:
Does this address #1689?

dmijatovic:
@cmeessen & @mvkma Ewan noticed we had an issue concerning profiles. Do you need more profiles? Currently I created two: `mail` and `data`. The `base` profile is not defined, on purpose, so the basic `docker compose up` command runs the `base` profile.

@cmeessen Can you help with further Makefile updates? It seems that development in docker does not work properly. In my test once I get BadGateway error and by running `make frontend-dev` the pages are loading very slowly for me. For me it running dev in docker seems not workable at the moment. Time ago it did worked so it might be related with the latest updates.

You are welcome to push additional changes as separate commit to this PR.

Makefile
```diff
@@ -30,7 +30,7 @@ export DGID
 # ----------------------------------------------------------------
 start: clean
 	docker compose build # build all services
-	docker compose up --scale data-generation=1 --scale scrapers=0 --detach
+	docker compose --profile data up --scale data-generation=1 --detach
```
ewan-escience: You can remove the `--scale` option.

Makefile
```diff
@@ -30,7 +30,7 @@ export DGID
 # ----------------------------------------------------------------
 start: clean
 	docker compose build # build all services
```
ewan-escience: You will need to build the `data` profile as well with `--profile data`. And edit the comment that indicated all services are built, since we don't build the `mail` services.

Makefile
```diff
@@ -44,7 +44,7 @@ install: clean
 	# All dependencies are installed. The data generation module is running in the background. You can now run `make dev' to start the application
 
 clean:
-	docker compose down --volumes
+	docker compose --profile mail --profile data down --volumes
```
ewan-escience: According to [the documentation](https://docs.docker.com/compose/how-tos/profiles/#start-multiple-profiles), you can use `--profile "*"` to apply the `down` command to all profiles.

Makefile
ewan-escience: Please edit this to remove `--scale` and use the profile.

ewan-escience: Please edit this to remove `--scale` and use the profile for the data-generation.

ewan-escience: Please edit this to remove `--scale` and use the profile.

ewan-escience: Unrelated, but I don't think this is needed anymore since we use health checks.

ewan-escience: Please edit this to remove `--scale` and use the profile for the data-generation.

ewan-escience (changes requested):
Nice work, but some more places that used `--scale` need to be updated, see the review comments. Also please update the following files:

- `data-generation/README.md`
- `documentation/docs/03-rsd-instance/01-getting-started.md`

And you can remove the comments regarding the docker compose version in the `Makefile`, since that version [is wrong](https://newreleases.io/project/github/docker/compose/release/1.28.0-rc1) and the actual version is over 5 years old.

mvkma:
> @cmeessen & @mvkma Ewan noticed we had an issue concerning profiles. Do you need more profiles? Currently I created two: `mail` and `data`. The `base` profile is not defined, on purpose, so the basic `docker compose up` command runs the `base` profile.

Nice, thank you for introducing this! Maybe it would make sense to also have a profile for the scrapers? In my local setup I've had this for a while because usually I don't need them.

dmijatovic:
> Nice, thank you for introducing this! Maybe it would make sense to also have a profile for the scrapers? In my local setup I've had this for a while because usually I don't need them.

@mvkma Would it work for you if we create scrapers profile and include it in `make start` in Makefile? In this case the `docker compose up` would not run scrapers but `make start` will. Or you prefer to cover both scenarios and add `make scraper`?

dalonsoa (commented):
Aside from @ewan-escience comments, everything seems to be working as stated. I think using profiles is much cleaner and easier to understand. Personally, I would manage as much as possible with profiles and not make, but I guess it has its use cases.

I have not tried to run the frontend in dev mode within docker, but had not used it before, so I wouldn't know what to expect there.

mvkma:
> @mvkma Would it work for you if we create scrapers profile and include it in `make start` in Makefile? In this case the `docker compose up` would not run scrapers but `make start` will. Or you prefer to cover both scenarios and add `make scraper`?

Yes, the addition of the scrapers profile would be great! Regarding the inclusion in `make start`: I think it makes sense what you're proposing, but I personally never use the `make` commands, so maybe my opinion here is not so valid. Maybe @cmeessen does and has some thoughts on it.

dmijatovic:
@ewan-escience, @cmeessen, @mvkma, @paulastock, @dalonsoa 

Thanks all for your feedback. I have implemented most of your remarks and updated PR description and instructions.  I would appreciate if you can confirm that this PR is workable for you or deliver any further desired changes.

I am also OK with creating a new issue to further consolidate Makefile, if needed. I am not sure who is using Makefile commands and which commands? I am using `make start` daily as main command to start everything for development. In addition, I start frontend manually via `npm run dev` from frontend folder using separate bash tab ( I prefer monitoring frontend logs in bash continuously).

ewan-escience (approved):
Looks good now. I pushed a commit to update the documentation. I don't use `make`, so maybe someone else should look at that in more detail.

dalonsoa (approved):
Looks good to me!

mvkma:
To me it also looks good! I also don't use `make` though.

dmijatovic:
@cmeessen I have changed `make e2e-tests` to reflect latest approach. I noticed that `make frontend-dev` and `make dev` do not work properly. For `make dev` I assume there is error in updating `env.local` file but I had no time to locate exact problem. I wonder if someone uses these features.

PR #1843, opened 2026-07-17
1836 modal styles

### LINKED ISSUE #1836: Improve basic modal styles to comply with existing RSD modals
Currently we add/copy styles to all modals we created in order to be in line with existing modals.

We should improve this approach and have modal styles be in line with existing modals by default.

There are two possibilities:
- Create RSD specific Modal components (in frontend/components/layout) as initaly suggested by Dusan.  
- Overwrite default modal styles in theme as suggested by Ewan.

I am in favor of Ewan's proposal but I would like to test if it works "properly". I expect that theme approach would work by default but also could be overwritten by passing "sx" directly to a new modal component. I think this gives us max flexibility and works directly out of box.

All other suggestions are welcome.

dmijatovic:
# Refactor modal styles

Closes #1836 

* Apply custom modal style in the theme to reduce amount of repeating styles in modal components.
* Remove custom styles from modals that are now part of "default" layout.
* Use standard hook for Save disabled state. There are few modal that use additional checks.
* Note! Since recently there is also custom async validation hook we should use for async input validation with react-hook-form. Example of use is in EditSoftwareBadgeModal component.
* Feedback modal changed to use standard modal colors and props.

How to test:
* `make start` to build and generate test data.
* Test all modals. There are quite a few, the most modals are in edit software, edit project and admin sections.

## Example feedback form (using theme colors)

image

PR Checklist:

*   [ ] Update documentation
*   [ ] Tests


ewan-escience (commented):
The "Send Feedback" dialog has a different style, is this on purpose? And the "Select organisation categories" doesn't seem to use these styles, as the tab order on the buttons is still broken.

dmijatovic:
> Select organisation categories

Well spotted! I have fixed tab order in Categories modal. Feedback modal is custom one designed by Jesse. I think we can leave it as is (as part of homepage design). I did changed tab order to be same as with other modals. @jmaassen What do you think?

dalonsoa (approved):
Looks good to me, and everything seems to work as expected. 

The feedback dialog is a bit of a headache for us. It is dark even tough we use the light theme, which causes a few accessibility issues for the contrast as there is no colour combination - within the Imperial palette - that works well with light and dark backgrounds. Not a huge deal, but a bit of inconsistency there that I thought worth mentioning.

dmijatovic:
> Looks good to me, and everything seems to work as expected.
> 
> The feedback dialog is a bit of a headache for us. It is dark even tough we use the light theme, which causes a few accessibility issues for the contrast as there is no colour combination - within the Imperial palette - that works well with light and dark backgrounds. Not a huge deal, but a bit of inconsistency there that I thought worth mentioning.

That's valid point. Let me see if I can "introduce" theme colors in feedback modal.

dmijatovic:
@dalonsoa Thanks for your feedback. We decided to convert Feedback modal to standard modal and solve customization and a11y issues.

@ewan-escience, @jmaassen Improved version is ready for tests.

ewan-escience (approved):

PR #1855, opened 2026-08-19
Fixed issue #1852, fixing the API documentation

### LINKED ISSUE #1852: Incorrect example in documentation on creating software through the access token API
In [the example of creating software through the access token API](http://localhost/documentation/API/rest-api/#example-2-create-a-new-software-entry), the Python code is incorrect. The header `Prefer: return=representation` is incompatible with creating an unpublished software entry (which is the case here, since the field `is_published` is omitted from the body, therefore having the default value `false`).

We should either delete the header or add the field `is_published` with a value of `true` to the body. In the latter case, the `curl` code should be adapted as well, to make the two code samples semantically equivalent.

aure44:
Fixing python API documentation

Fixes #1852 

*    Removed the Prefer: return=representation header from the Python example, which was not present in the equivalent curl request and was producing an error. The two examples now perform the same call.

How to test:

* Run the Python snippet

PR Checklist:

*   [x] Update documentation
*   [ ] Tests

### DIFF
--- a/documentation/docs/02-API/01-rest-api.md
+++ b/documentation/docs/02-API/01-rest-api.md
@@ -96,8 +96,7 @@ print(response.json())
         data = {"slug": "test-software", "brand_name": "TEST-Software", "description": "My new software entry"}
         headers = {
             'Authorization': f'Bearer {accessToken}',
-            'Content-Type': "application/json",
-            "Prefer": "return=representation"
+            'Content-Type': "application/json"
             }
 
         response = requests.post(url, json=data, headers=headers)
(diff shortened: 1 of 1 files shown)


documentation/docs/02-API/01-rest-api.md
ewan-escience:
```suggestion
        print(response.text)
```

ewan-escience (changes requested):
Thanks for the PR! The `json()` function call fails because the response is empty, can you change it as indicated?

documentation/docs/02-API/01-rest-api.md
ewan-escience: And remove the old line (GitHub is not cooperating).

aure44: Hello Ewan, the change has been made

ewan-escience (approved):
Thanks for your contribution!

